PR #44, opened 2020-10-02
Add trap mechanism to handle term signals. Fixes #28

### LINKED ISSUE #28: (Intermediate) If running the docker image in foreground, sending a kill signal doesn't stop the container
To recreate the problem that exists today:
- run `docker pull supertokens/supertokens-mysql:2.4`
- Start the docker image in foreground: `docker run -p 3567:3567 supertokens/supertokens-mysql:2.4`
- Wait for it to start.. It should display `Started SuperTokens on 0.0.0.0:3567 with PID: <some PID>`
- Now send the kill signal (Ctrl + C / Z on a Mac)

At this point, the Java process in the docker process should stop and the docker process itself should quit. However, that does not happen.

The docker file for the above image can be found in [this repo](https://github.com/supertokens/supertokens-docker-mysql). Please feel free to fork that as well.

As some extra info:
- if not using docker (manually installing it on your system), sending the kill signal does stop the running process.
- In the Java code (this repo), in [Main.java](https://github.com/supertokens/supertokens-core/blob/master/src/main/java/io/supertokens/Main.java), there is [a function](https://github.com/supertokens/supertokens-core/blob/master/src/main/java/io/supertokens/Main.java#L354) that registers a shutdown hook.

Fabricio20:
This commit adds a trap mechanism around the exec call to pass the signals received by bash to a child process.

Due to added complexity and for better legibility, I've opted to move the source for both scripts from java to a resource.

These changes are based on [sigterm propagation](http://veithen.io/2014/11/16/sigterm-propagation.html).

I was not able to test the changes in their full extent, due to compilation issues on this project, I did however successfully test the change by overriding via volume mounts in docker. Let me know if this breaks windows compatibility or doesn't work after a full build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,7 @@ If you think this is a project you could use in the future, please :star2: this 
 <td align="center"><a href="https://github.com/UbadahJ"><img src="https://avatars1.githubusercontent.com/u/26687928?s=460&u=ae1d3ae5fad6e4cfa71809f8ce4a99429321dcaf&v=4" width="100px;" alt=""/><br /><sub><b>LordChadiwala</b></sub></a></td>
 <td align="center"><a href="https://github.com/LuizDoPc"><img src="https://avatars0.githubusercontent.com/u/20651653?s=460&u=d673e5357da83e446311831efe107e695d3ef875&v=4" width="100px;" alt=""/><br /><sub><b>Luiz Soares</b></sub></a></td>
 <td align="center"><a href="https://github.com/sudiptog81"><img src="https://avatars0.githubusercontent.com/u/11232940?s=460&u=07b4989ae4c43e43f35730d7f8d59631f5ed933c&v=4" width="100px;" alt=""/><br /><sub><b>Sudipto Ghosh</b></sub></a></td>
+<td align="center"><a href="https://github.com/Fabricio20"><img src="https://avatars1.githubusercontent.com/u/7545720?s=400&v=4" width="100px;" alt=""/><br /><sub><b>Fabricio20</b></sub></a></td>
   </tr>
 </table>
 
--- a/README.md
+++ b/README.md
@@ -188,6 +188,9 @@ If you think this is a project you could use in the future, please :star2: this 
 <td align="center"><a href="https://github.com/arnxv0"><img src="https://avatars2.githubusercontent.com/u/57629464?s=460&u=5f0cca1aed9fabb38bea74df73ed99dfcfec2f26&v=4" width="100px;" alt=""/><br /><sub><b>Arnav Dewan</b></sub></a></td>
 <td align="center"><a href="https://github.com/NkxxkN"><img src="https://avatars1.githubusercontent.com/u/5072452?s=460&u=eda6b25b674d20e3389bf19a0619d6e4c1e46670&v=4" width="100px;" alt=""/><br /><sub><b>NkxxkN</b></sub></a></td>
   </tr>
+  <tr>
+<td align="center"><a href="https://github.com/UbadahJ"><img src="https://avatars1.githubusercontent.com/u/26687928?s=460&u=ae1d3ae5fad6e4cfa71809f8ce4a99429321dcaf&v=4" width="100px;" alt=""/><br /><sub><b>LordChadiwala</b></sub></a></td>
+  </tr>
 </table>
 
 ## Contributing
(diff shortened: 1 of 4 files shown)


rishabhpoddar:
@Fabricio20 thanks for the contribution. Will the `install-linux.sh` and `install-windows.bat` files get bundled in the JAR itself, or will these files also need to be distributed along with the JAR?

Fabricio20:
Hey there, these two files come bundled in the final jar.

rishabhpoddar:
@bhumilsarvaiya, could you please review this PR.

You need to build a docker image based on the forked core. Instructions on how to build a docker image can be found here: https://github.com/supertokens/supertokens-core/wiki/Building-from-source#creating-a-docker-image (here it says using mysql, but you can follow the same steps with sqlite as well)

Once the image is created, you need to test:
- Starting the docker image in background and querying it.. it should work.
- Starting the docker image in foregrdoun (with `-d`) and checking it works, then pressing ctrl + c and seeing that it quite smoothly.
- SSH into the docker image (which is not running using `docker run -it supertokens bash`). Then run `supertokens --help` and see the output is shown properly and that you gain control of the shell after the output is done.
- SSH into the docker image (which is not running using `docker run -it supertokens bash`). Then start supertokens using `supertokens start` and see that it starts in the background and that you can run `supertokens list` to see it.
- SSH into the docker image (which is not running using `docker run -it supertokens bash`). Then start it in the foreground using `supertokens start --foreground` and see it run. Then pressing ctrl + c should kill it.

Then independent to the docker image, install SuperTokens on your machine (using the forked core) using the instruction found here: https://github.com/supertokens/supertokens-core/wiki/Building-from-source#installing-manually

After installation:
- Run `supertokens start` to see it start in the background and then you can run `supertokens list` to see it and `supertokens stop` to stop it.
- Run `supertokens start --foreground` to see it run in the foreground, and then pressing ctrl + c should kill the process.
- Run `supertokens --help` to see the output of that and there should be nothing weird or something like `grep: /proc/1/cgroup: No such file or directory` being printed out in the console.

bhumilsarvaiya:
Tested the following:
- [ ] Starting the docker image in background and querying it.. it should work.
- [ ] Starting the docker image in foreground (without -d) and checking it works, then pressing ctrl + c and seeing that it quite smoothly.
- [ ] SSH into the docker image (which is not running using docker run -it supertokens bash). Then run supertokens --help and see the output is shown properly and that you gain control of the shell after the output is done.
- [ ] SSH into the docker image (which is not running using docker run -it supertokens bash). Then start supertokens using supertokens start and see that it starts in the background and that you can run supertokens list to see it.
- [ ] SSH into the docker image (which is not running using docker run -it supertokens bash). Then start it in the foreground using supertokens start --foreground and see it run. Then pressing ctrl + c should kill it.
---
- [ ] Run supertokens start to see it start in the background and then you can run supertokens list to see it and supertokens stop to stop it.
- [ ] Run supertokens start --foreground to see it run in the foreground, and then pressing ctrl + c should kill the process.
- [ ] Run supertokens --help to see the output of that and there should be nothing weird or something like grep: /proc/1/cgroup: No such file or directory being printed out in the console.
---
Hey @Fabricio20, great work! All the docker tests are running correctly! ✌️ 
For the remaining of the supertokens test (i.e. when installed manually), I kept getting following message printed out whenever I tried to execute any cli command (the commands were getting executed correctly but this message was always getting printed):
```bash
grep: /proc/1/cgroup: No such file or directory
```
I was getting this issue on `macOS Catalina 10.15.7`.

I got it fixed by changing the `line 5` from file [`cli/src/main/resources/install-linux.sh`](https://github.com/Fabricio20/supertokens-core/blob/master/cli/src/main/resources/install-linux.sh):
```bash
if grep docker /proc/1/cgroup -qa; then
```
to
```bash
if [ -f /proc/1/cgroup ] && grep docker /proc/1/cgroup -qa; then
```
All the tests were passing after the above change.

---
For manually installed supertokens, if the process is started in foreground, the supertokens process seems to end but it doesn't seem to return back to terminal. I tried doing it couple of times but the issue remains. When I tried to exit the program using either `Ctrl + c` or `Ctrl + d`, this is what kept happening:
Screenshot 2020-10-07 at 6 57 54 PM

rishabhpoddar:
@Fabrihcio20 , @bhumilsarvaiya , amazing work! Thank you.

@bhumilsarvaiya , the last problem you mentioned about running SuperTokens using `--foreground` is a [different issue](https://github.com/supertokens/supertokens-core/issues/49), and has nothing to do with this issue or the changes made for this issue.

rishabhpoddar:
@Fabricio20 and @bhumilsarvaiya amazing work! Thank you.

The last problem you mentioned about running SuperTokens using --foreground is a different issue, and has nothing to do with this issue or the changes made for this issue.

PR #45, opened 2020-10-03
Correct test for revokeSessionWithBlacklisting to throw unauthorized …

### LINKED ISSUE #31: Add JWT blacklisting feature
- This should be switched off by default.
- Implement it by checking if the associated session handle exists in the database

NkxxkN:
Fixes #31 

> @NkxxkN , is there is a test where blacklisting is switched off, and we check that even if the session is revoked, the access token still works?

Yes there is [here](https://github.com/supertokens/supertokens-core/blob/master/src/test/java/io/supertokens/test/SessionTest2.java#L199-L247)

But the test `revokeSessionWithBlacklist` was not setting the `blacklisting` config, hence doing the same as without blacklisting.

I made sure that it was the only test `...WithBlacklisting` which didn't behave properly.


rishabhpoddar:
Thanks for the fix!

Issue #255, opened 2021-06-03
Add ability to get userId from sessionHandle


rishabhpoddar:
Something like `getUserId(sessionHandle)`

## TODOs:
- [x] update swagger spec & cdi md file
- [x] update node SDK & add tests (basic)
- [ ] update docs
    - [ ] recipe docs for sessions and others
    - [ ] nodeJS docs

rishabhpoddar:
Actually now that I think of this again, we will have three APIs for querying based on session:
- To get JWT payload
- To get Session data
- To get userId

We should just combine them into one API which returns all these three info (from the core's perspective), and then in the node side (user's point of view), we can still have three functions - one for each type of info.

nkshah2:
@rishabhpoddar Just to clarify, we still want to leave these APIs intact
- `/recipe/session/data GET`
- `/recipe/jwt/data GET`

to avoid backward compatibility problems and simple introduce a new one that combines all 3 functionalities?

Also if we are leaving them intact, do we also want to have a separate one for just userId to keep things consistent?

rishabhpoddar:
> to avoid backward compatibility problems and simple introduce a new one that combines all 3 functionalities?
Yes. And also see what other makes sense to return apart from session data, jwt payload and userId

> Also if we are leaving them intact, do we also want to have a separate one for just userId to keep things consistent?
No need. We will mark those other two APIs as deprecated in the Spec. 

---------
You should start by issuing a PR for a new CDI spec for this file: https://github.com/supertokens/core-driver-interface/blob/master/v2.8.0.md

You should also update the issue here: https://github.com/supertokens/core-driver-interface/issues/8 summarising the change you have made.

PR #266, opened 2021-06-26
Add the ability to get userId and session information using sessionHandle

nkshah2:
This pull request adds a way to fetch userId for a given sessionHandle. Since the functionality from supertokens-core's perspective is the same as fetching JWT and sessionData, this PR also combines the functionality of `/recipe/jwt/data GET` and `/recipe/session/data GET` into a single API `/recipe/session GET`. This API returns:
- userId
- userDataInDatabase
- userDataInJWT
- expiry
- timeCreated

Changes include:
- Adding `/recipe/session GET` endpoint
- New function to fetch data in the session recipe source

## Link to a github issue 

https://github.com/supertokens/supertokens-core/issues/255

### Have you read the [Contributing Guidelines on pull requests](https://github.com/supertokens/supertokens-core/blob/master/CONTRIBUTING.md#pull-request)?

Yes

## Test Plan

Changes were verified by adding tests for new data fetch functions and changes to the `/recipe/session` API. All current and new tests pass.

Screenshot 2021-06-26 at 4 55 10 PM

## Documentation changes (checked items are already done)

- [x] Add spec to core-driver-interface (changes made to https://github.com/supertokens/core-driver-interface/blob/master/v2.8.0.md)
- [ ] Add spec to swagger


src/main/java/io/supertokens/webserver/api/session/SessionAPI.java
```diff
@@ -83,4 +85,32 @@ protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws I
...
+            SessionInfo sessionInfo = Session.getSession(main, sessionHandle);
+
+            JsonObject result = new JsonObject();
+
+            result.addProperty("status", "OK");
+            result.add("userDataInDatabase", sessionInfo.userDataInDatabase);
+            result.add("userDataInJWT", sessionInfo.userDataInJWT);
+            result.addProperty("userId", sessionInfo.userId);
+            result.addProperty("expiry", sessionInfo.expiry);
+            result.addProperty("timeCreated", sessionInfo.timeCreated);
```
rishabhpoddar, line 103: You can use `transient` keyword in the `SessionInfo` class for properties that you don't want to be included in the JSON. With that, you can directly convert `sessionInfo` instance to JSON.

rishabhpoddar (changes requested):

rishabhpoddar:
@nkshah2 you should also mark the other APIs that this replaces as `@deprecated`. You can do this by adding this annotation to:
- The class that implements those APIs
- The `doGet` / `doPost` functions themselves within those classes
- The helper functions inside the `Session.java` class
- The relevant functions (if any) in the plugin-interface project.

Essentially, the only warning for deprecated usage should come from the `Webserver.java` class

src/test/java/io/supertokens/test/session/SessionGetJWTDataTest.java
```diff
@@ -0,0 +1,195 @@
...
+
+
+        SessionInformationHolder sessionInfo = Session
+                .createNewSession(process.getProcess(), userId, userDataInJWT, userDataInDatabase, false);
+
+        //let it expire, remove from db
+
+        Thread.sleep(2000);
+        String[] sessionHandles = {sessionInfo.session.handle};
+        Session.revokeSessionUsingSessionHandles(process.getProcess(), sessionHandles);
```
rishabhpoddar, line 181: Should you need to call revoke at all? If the refresh token has also expired, the session should be as good as revoed.. So maybe add another test that does the same thing but does not call revoke (this may require changes to your logic for fetching session info to check the expiry time)

nkshah2:
@rishabhpoddar Since I would have to make the class properties in the plugin-interface transient, how do we manage the PRs? Should I raise another for plugin-interface before we merge this one?

src/test/java/io/supertokens/test/session/api/SessionGetAPIJWTTest.java
```diff
@@ -0,0 +1,164 @@
...
+        TestingProcessManager.TestingProcess process = TestingProcessManager.start(args);
+        assertNotNull(process.checkOrWaitForEvent(ProcessState.PROCESS_STATE.STARTED));
+
+        //Get Request input errors
+        //null is sent in parameters
+        try {
+            io.supertokens.test.httpRequest.HttpRequest
+                    .sendGETRequest(process.getProcess(), "", "http://localhost:3567/recipe/session", null, 1000,
+                            1000,
+                            null, Utils.getCdiVersionLatestForTests(), Session.RECIPE_ID);
```
rishabhpoddar, line 65: Need to change to getting 2_8 CDI explicitly - no action from your side required at the moment.

src/test/java/io/supertokens/test/session/api/SessionGetAPIJWTTest.java
```diff
@@ -0,0 +1,164 @@
...
+
+        //typo in parameter
+        HashMap<String, String> map = new HashMap<>();
+        map.put("sessiondle", "");
+
+        try {
+            io.supertokens.test.httpRequest.HttpRequest
+                    .sendGETRequest(process.getProcess(), "", "http://localhost:3567/recipe/session", map, 1000,
+                            1000,
+                            null, Utils.getCdiVersionLatestForTests(), Session.RECIPE_ID);
```
rishabhpoddar, line 81: Use `"session"` instead of `Session.RECIPE_ID`

src/test/java/io/supertokens/test/session/api/SessionGetAPIJWTTest.java
```diff
@@ -0,0 +1,164 @@
...
+                sessionBody.get("userDataInDatabase").getAsJsonObject());
+
+        assertEquals(response.get("userDataInJWT").getAsJsonObject(),
+                sessionBody.get("userDataInJWT").getAsJsonObject());
+
+        assertEquals(response.get("userId").getAsString(),
+                sessionBody.get("userId").getAsString());
+
+        assertNotNull(response.get("expiry"));
+        assertNotNull(response.get("timeCreated"));
```
rishabhpoddar, line 162: May want to check that the type of these are `numbers`

rishabhpoddar:
@nkshah2 yes. You can raise a PR to the branch that you are currently using for plugin-interface.

rishabhpoddar (changes requested):

rishabhpoddar:
@nkshah2 sorry, please make the PR to branch 2.8 of the plugin-interface repo. If there are conflicts, let them be.. i'll resolve them.

src/test/java/io/supertokens/test/session/SessionGetJWTDataTest.java
nkshah2: Yeah I did think of that but I saw it being done in another test so stuck with it. Adding a different test for this is a good idea, on it

src/test/java/io/supertokens/test/session/api/SessionGetAPIJWTTest.java
nkshah2: Done

src/test/java/io/supertokens/test/session/api/SessionGetAPIJWTTest.java
nkshah2: Done

src/test/java/io/supertokens/test/session/SessionGetJWTDataTest.java
nkshah2: Done

nkshah2:
@rishabhpoddar https://github.com/supertokens/supertokens-plugin-interface/pull/17 PR for plugin-interface

nkshah2:
Done with all the changes, all existing and added tests pass
Screenshot 2021-06-27 at 1 47 31 PM

src/main/java/io/supertokens/webserver/api/session/JWTDataAPI.java
```diff
@@ -42,7 +43,7 @@ public String getPath() {
         return "/recipe/jwt/data";
     }
 
-    @Override
+    @Override @Deprecated
```
rishabhpoddar, line 46: The `put` API isn't deprecated right? Just the `GET` API... so you would also have to remove the `@Deprecated` from this class.

src/main/java/io/supertokens/webserver/api/session/SessionDataAPI.java
```diff
@@ -64,7 +65,7 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws IO
         }
     }
 
-    @Override
+    @Override @Deprecated
```
rishabhpoddar, line 68: Put API isn't deprecated.

src/main/java/io/supertokens/session/Session.java
```diff
@@ -554,7 +557,8 @@ public static JsonObject getJWTData(Main main, String sessionHandle)
         io.supertokens.pluginInterface.session.SessionInfo session = StorageLayer.getSessionStorage(main)
                 .getSession(sessionHandle);
 
-        if (session == null) {
+        // If there is no session, or session is expired
+        if (session == null || session.expiry <= new Date().getTime()) {
```
rishabhpoddar, line 561: use `System.currentTimeMillis()` instead - to be consisten with other places where we generate curr time.

rishabhpoddar: Also, if you remove `|| session.expiry <= new Date().getTime()`, i believe that one of your tests will fail right - if not, then please look into it to make sure that at least one of them do?

rishabhpoddar (changes requested):

src/main/java/io/supertokens/session/Session.java
nkshah2: Without the `|| session.expiry <= new Date().getTime()` the test that checks for refresh token expired without calling revokeSession fails

nkshah2: Done

src/main/java/io/supertokens/webserver/api/session/SessionDataAPI.java
nkshah2: Done

src/main/java/io/supertokens/webserver/api/session/JWTDataAPI.java
nkshah2: Yeah youre right, shouldve noticed this. Done

rishabhpoddar (approved):

rishabhpoddar:
@nkshah2 I had commented in the plugin-interface PR about keeping `sessionHandle` as part of the JSON response. Do you think that that's a good idea? If yes, then the API tests here will need to change to account for that, else nothing more is needed here.

nkshah2:
@rishabhpoddar it doesnt hurt to keep it as part of the response, at best it can be used to verify and at worst it can be safely ignored. Ill make the changes and update the plugin-interface PR. Ill raise another one for the spec on core-driver-interface too

rishabhpoddar:
@nkshah2 please make a change to the 3.5 version of the core as I have merged this already..

Issue #277, opened 2021-07-12
Fetch ThirdParty users based on email


dulowski-marek:
### Goal
Developer should be able to fetch multiple users associated with one email.

rishabhpoddar:
- I think you should have a look at the CDI spec first: https://github.com/supertokens/core-driver-interface/blob/master/v2.8.0.md. This API spec is already defined in there.

> What permissions are required to execute this action?

We don't have roles and permissions.. just an optional API key that the core has.

rishabhpoddar:
@dulowski-marek im not sure what to review for this.. the core changes for this have already been merged. The next part of this task is a change in the node SDK, the PR for which is missing..

PR #278, opened 2021-07-17
277: Add GetThirdPartyUsersByEmail API

dulowski-marek:
## Related

- https://github.com/supertokens/supertokens-core/issues/277
- (blocked by) https://github.com/supertokens/supertokens-plugin-interface/pull/18

## Test Plan

1. Create two users with same email `john.doe@example.com`
2. Query the GetThirdPartyUsersByEmail endpoint with email = `john.doe@example.com`
3. Expect response with 2 users with email = `john.doe@example.com`

## Problems
1. The CDI spec specifies `GET /recipe/user` endpoint path https://github.com/supertokens/core-driver-interface/blob/master/v2.8.0.md#recipeuser-get-1, but it doesn't make sense, because the response could then contain either one user or multiple users (when querying by `userId`/thirdparty ids vs `email`). It would make most sense to expose it as `GET /recipe/users?by=email&email=...`, but the endpoint has been deprecated. Therefore new endpoint has been made, but I'm absolutely open to changing its path for better.

2. The endpoint test where users are created fails due to `../config.yaml` not being found. I'm not sure what the issue is.


rishabhpoddar:
Add one more test:
- No third party users exist, and querying by email yields an empty array.

> Therefore new endpoint has been made, but I'm absolutely open to changing its path for better.

Yes. Please make a new endpoint

> The endpoint test where users are created fails due to ../config.yaml not being found. I'm not sure what the issue is.

If you are running via intellij, make sure to use the `./startTestingEnv --wait` command before.. If this doesn't work, I can discuss over a call.

src/main/java/io/supertokens/inmemorydb/queries/ThirdPartyQueries.java
```diff
@@ -172,19 +162,40 @@ public static UserInfo getUserInfoUsingId_Transaction(Start start, Connection co
...
         }
     }
 
+    public static UserInfo[] getThirdPartyUsersByEmail(Start start, @Nonnull String email)
+        throws SQLException, StorageQueryException {
+
+        String sqlQuery = "SELECT user_id, third_party_id, third_party_user_id, email, time_joined FROM " +
+                Config.getConfig(start).getThirdPartyUsersTable() +
+                " WHERE email = ?" +
+                " ORDER BY user_id DESC";
```
rishabhpoddar, line 176: I don't think you need to add `ORDER BY user_id DESC` to the query.

src/main/java/io/supertokens/inmemorydb/queries/ThirdPartyQueries.java
```diff
@@ -172,19 +162,40 @@ public static UserInfo getUserInfoUsingId_Transaction(Start start, Connection co
...
+
+        if (email.equals("*")) {
```
rishabhpoddar, line 178: Input sanitization should happen in the API layer, not here. There, you should check that the input email correctly matches the format of an email. Anyway, with prepared statements using `*` as email will just yield an empty result.

src/main/java/io/supertokens/thirdparty/ThirdParty.java
```diff
@@ -159,6 +161,10 @@ public static UserPaginationContainer getUsers(Main main,
         return new UserPaginationContainer(resultUsers, nextPaginationToken);
     }
 
+    public static UserInfo[] getUsersByEmail(Main main, @Nonnull GetUsersByEmailQuery query) throws StorageQueryException {
```
rishabhpoddar, line 164: This function should just take `Main main, @Nonnull String email` as the input. No need to create an abstraction for `GetUsersByEmailQuery`... keep it simple :)

src/main/java/io/supertokens/webserver/api/thirdparty/getUsersByEmail/GetUsersByEmailQuery.java
```diff
@@ -0,0 +1,42 @@
+/*
```
rishabhpoddar, line 1: I don't think this abstraction is required.. no other API is made like this, so I would insist that we stick to existing practices. If you disagree, can you please list down **very clear** benefits of doing something like this, and then if those benefits are indeed much better, then we can (overtime), change all APIs to have this level of abstraction.

src/main/java/io/supertokens/webserver/api/thirdparty/getUsersByEmail/GetUsersByEmailAPI.java
```diff
@@ -0,0 +1,64 @@
...
+import io.supertokens.pluginInterface.thirdparty.UserInfo;
+import io.supertokens.thirdparty.ThirdParty;
+import io.supertokens.webserver.WebserverAPI;
+
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+public class GetUsersByEmailAPI extends WebserverAPI {
```
rishabhpoddar, line 34: You need to add this to `io.supertokens.webserver.Webserver` > `setupRoutes` function. Otherwise it's not exposed..

src/test/java/io/supertokens/test/GetUsersByEmailAPITest2_8.java
```diff
@@ -0,0 +1,69 @@
...
+
+        // given
+        JsonObject signUpResponse = Utils.signInUpRequest_2_7(process, "john.doe@example.com", true, "mockThirdParty", "mockThirdPartyUserId");
+        assertEquals(signUpResponse.get("status").getAsString(), "OK");
+        assertEquals(signUpResponse.entrySet().size(), 2);
+
+        JsonObject signUpUser = signUpResponse.get("user").getAsJsonObject();
+        assertEquals(signUpUser.get("email").getAsString(), "random@gmail.com");
+        assertNotNull(signUpUser.get("id"));
+    }
```
rishabhpoddar, line 68: You need to kill the process as well. Please see how other tests are done.

src/test/java/io/supertokens/test/GetUsersByEmailAPITest2_8.java
```diff
@@ -0,0 +1,69 @@
...
+        Utils.afterTesting();
+    }
+
+    @Before
+    public void beforeEach() {
+        Utils.reset();
+    }
+
+    @Test
+    public void testReturnTwoUsersWithSameEmail() throws Exception {
```
rishabhpoddar, line 49: Not sure if this is testing your API in any way.. Or maybe this is incomplete?

src/test/java/io/supertokens/test/GetUsersByEmailQueryTest.java
```diff
@@ -0,0 +1,55 @@
+/*
```
rishabhpoddar, line 1: Not sure why this file exists.. please discuss with me how testing needs to be done. If you have a better method that is clearly much, much better, then we can go with that. Else, please stick to what we currently do.

rishabhpoddar (changes requested):

src/main/java/io/supertokens/webserver/api/thirdparty/getUsersByEmail/GetUsersByEmailQuery.java
dulowski-marek: ### Current state
Currently we have 3 layers of application: APIs, Business-logic Functions and Storage Functions.
Imagine a scenario where we decide to add a query parameter to the endpoint. 

All the incoming data is passed as positional argument primitives to functions in Business-logic Functions and Storage Functions, like so:
`(receives a request) -> EmailPassword.findUsers(emails, limit, orderBy, ...) -> StorageLayer.getEmailPassword(...).findUsers(emails, limit, orderBy, ...)`. 

#### Consequence
If we decide to add a parameter, each of these functions has to change signature. So, in more high level terms, change in API causes change of signatures in layers of application that should be isolated from API changes.

### Using Query/Command classes as Argument Objects
Now think about doing such query/command object, which is just an Argument Object (as per uncle bob's Clean Code):
```java
public class FindUsersQuery {
    public List<String> getEmails() {}
    public Optional<Integer> getLimit() {}
    ...
}
```

If we use it, change in API won't cause the change in business logic. We may rename some parameters, introduce new optional ones and it won't cause a cascade of changes in signatures. Hence, we won't have to provide multiple Business-logic Functions and Storage Functions overloads to account for different API variants. This makes the API way clearer.

### Unnecessary abstraction
The fact that we don't yet know what will change in API doesn't deem this an unnecessary abstraction. If inside layers (think Business-logic Functions and Storage Functions) are separated from the API, passing this API input as positional arguments is a weak spot (because it forces multiple layers to change signatures or provide overloads).

### Why I used it
Because I think it's a low-cost, medium-value solution. The parameters for given API call anyway make sense when grouped, not separately.

rishabhpoddar: Understood. So this Object would carry all the way through the storage layer right until the params are used in the actual queries?

If not, then I can argue that the benefit this gives is limited to none (cause then only this layer: `(receives a request) -> EmailPassword.findUsers(emails, limit, orderBy, ...)` would be unaffected and StorageLayer functions would have to change anyway)

If yes, then this object would need to be defined in the plugin interface project as the storage layer only depends on the plugin interface and not the core project. Does it make sense to add a new class for each function exposed via the plugin interface?

-------
I assume your final aim is to make is to that if the API's input change, all that needs to change (for most cases) is the input validation in this class and the final SQL query for that API?

dulowski-marek: > So this Object would carry all the way through the storage layer right until the params are used in the actual queries?

That's correct. It prevents from violating open-closed principle all the way down through the application (which happens if signatures need to change). So I would say unless we have a really good reason to do it, I would stick to those query/command objects (regardless of whether validation happens there, main aim is the architectural correctness).

> If yes, then this object would need to be defined in the plugin interface project as the storage layer only depends on the plugin interface and not the core project.

That's true, I noticed it as well. Is there a reason core depends on plugin-interface and not the other way around? To me it would make more sense if plugin-interface classes depended on interfaces defined in core.

> Does it make sense to add a new class for each function exposed via the plugin interface?

This is my first thought. Each function exposed via the plugin interface handles separate usecase. Doing that we can achieve similar level of separation as we have across recipes, which is great over time for managing such APIs.

> I assume your final aim is to make is to that if the API's input change, all that needs to change (for most cases) is the input validation in this class and the final SQL query for that API?

Yes, this way we don't break open-closed principle just to maintain simplicity.

rishabhpoddar: > That's true, I noticed it as well. Is there a reason core depends on plugin-interface and not the other way around? To me it would make more sense if plugin-interface classes depended on interfaces defined in core, but I might be missing something.

The plugin-interface is meant for communication between the core and the db plugins. Since it's just an interface, it should not depend on the "thing" that is using the interface or implementing the interface.

----------
One point to note that the plugin interface is not the same as the API interface.. By doing this and adding these new classes which are declared in the plugin-interface and created in the API classes is a bit weird.. If the API changes, then we wouldn't want the plugin-interface class to change (unless necessary).

src/main/java/io/supertokens/thirdparty/ThirdParty.java
```diff
@@ -158,6 +162,18 @@ public static UserPaginationContainer getUsers(Main main,
         return new UserPaginationContainer(resultUsers, nextPaginationToken);
     }
 
+    public static UserInfo[] getUsersByEmail(Main main, String email) throws StorageQueryException, InvalidInputException {
+        if (email == null || email.equals("")) {
```
rishabhpoddar, line 166: - As per our current API design, validation of input happens in the API and not here.. But I'm OK with this. 
- If you want email to not be null, you should mark the input param as @Nonnull and not have to check here again.
- Also, if you are checking `EmailValidator.isValid` below, why do you need to check if it's not an empty string here?

src/main/java/io/supertokens/utils/EmailValidator.java
```diff
@@ -0,0 +1,34 @@
...
+ *    License for the specific language governing permissions and limitations
+ *    under the License.
+ */
+
+package io.supertokens.utils;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public abstract class EmailValidator {
```
rishabhpoddar, line 22: You don't need a new class for this.. It can simply be a static function in the `utils.java` class.. No need to create extra classes please.

src/main/java/io/supertokens/webserver/api/thirdparty/getUsersByEmail/GetUsersByEmailAPI.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Override
+    public String getPath() {
+        return "/recipe/users/by-email";
+    }
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOException, ServletException {
+        try {
+            String email = req.getParameter("email");
```
rishabhpoddar, line 49: - I would prefer if you do email validation here. 
- We have functions that read from the request object and enforce that they are not null. Please see other GET APIs for this.

src/main/java/io/supertokens/webserver/api/thirdparty/getUsersByEmail/GetUsersByEmailAPI.java
```diff
@@ -0,0 +1,66 @@
...
+import java.io.IOException;
+
+public class GetUsersByEmailAPI extends WebserverAPI {
+    public GetUsersByEmailAPI(Main main) {
+        super(main, RECIPE_ID.THIRD_PARTY.toString());
+    }
+
+    @Override
+    public String getPath() {
+        return "/recipe/users/by-email";
```
rishabhpoddar, line 43: Have you updated the swagger spec with this already?

src/test/java/io/supertokens/test/GetUsersByEmailAPITest2_8.java
```diff
@@ -0,0 +1,150 @@
...
+
+        // when
+        HashMap<String, String> query = new HashMap<>();
+        query.put("email", "john.doe@example.com");
+
+        JsonObject response = io.supertokens.test.httpRequest.HttpRequest
+                .sendGETRequest(process.getProcess(), "",
+                        "http://localhost:3567/recipe/users/by-email", query, 1000,
+                        1000,
+                        null, "2.8", RECIPE_ID.THIRD_PARTY.toString());
```
rishabhpoddar, line 87: Do not hard code "2.8". Please see other tests on how this is done...

src/test/java/io/supertokens/test/GetUsersByEmailAPITest2_8.java
```diff
@@ -0,0 +1,150 @@
...
+        } catch (io.supertokens.test.httpRequest.HttpResponseException e) {
+            assertEquals(400, e.statusCode);
+            assertEquals("Http error. Status Code: 400. Message: email cannot be empty", e.getMessage());
+        }
+
+        // when
+        try {
+            Map<String, String> invalidQuery = new HashMap<>();
+
+            invalidQuery.put("email", "");
```
rishabhpoddar, line 128: Since you are checking for email regex validation, you should also add that check?

src/test/java/io/supertokens/test/GetUsersByEmailAPITest2_8.java
```diff
@@ -0,0 +1,150 @@
+/*
```
rishabhpoddar, line 1: You also need a test in ThirdParty.java to test the raw functions properly:
- If you use the get users by email function with no users, it should return an empty array
- If you provide an email that doesn't exist (but there are users), it should return an empty array
- If you have multiple users with different emails, and you use one email, it should return the relevant subset of users.

Again, please see how other functions + APIs are tested..

rishabhpoddar (changes requested):

src/test/java/io/supertokens/test/GetUsersByEmailAPITest2_8.java
dulowski-marek: It's an edge case. If I didn't test that here, I would anyway test regex if it matched an empty string.

src/test/java/io/supertokens/test/GetUsersByEmailAPITest2_8.java
dulowski-marek: > You also need a test in ThirdParty.java to test the raw functions properly:

I don't know what you mean by raw functions. Do you mean to unit test them? I'm all for it, but there's no other test I could find that did that. 

> * If you use the get users by email function with no users, it should return an empty array
> * If you provide an email that doesn't exist (but there are users), it should return an empty array
> * If you have multiple users with different emails, and you use one email, it should return the relevant subset of users.

This makes sense.

src/main/java/io/supertokens/webserver/api/thirdparty/getUsersByEmail/GetUsersByEmailAPI.java
dulowski-marek: I did that

src/test/java/io/supertokens/test/GetUsersByEmailAPITest2_8.java
dulowski-marek: Changed to use a function.

src/main/java/io/supertokens/thirdparty/ThirdParty.java
dulowski-marek: > * As per our current API design, validation of input happens in the API and not here.. But I'm OK with this.

Good, I'm certain that validation not happening in the API layer is a good design choice (isolation from API layer changes and reduced duplication).

> * If you want email to not be null, you should mark the input param as @nonnull and not have to check here again.

That's why there's `@Nonnull` on storage layer parameter. I don't want to move this bit of validation logic "up", to an entity that calls `getUsersByEmail`. This would lead to unnecessary duplication.

> * Also, if you are checking `EmailValidator.isValid` below, why do you need to check if it's not an empty string here?

I will move that to `EmailValidatior.isValid`, it makes more sense.

src/main/java/io/supertokens/utils/EmailValidator.java
dulowski-marek: Just as we discussed, `Utils.java` is bad. Everything is mixed there and it's harder to find various utilities (since they are namespaced). I think we will be better with `EmailValidator.java`: no effort to make and test it, higher clarity and no dependency magnets (as `Utils.java` is).

src/main/java/io/supertokens/webserver/api/thirdparty/getUsersByEmail/GetUsersByEmailAPI.java
dulowski-marek: > * I would prefer if you do email validation here.

For consistency I'd agree, but there's a good architectural reason not to (as in comment above).

> * We have functions that read from the request object and enforce that they are not null. Please see other GET APIs for this.

I am aware, although this executes the validation logic in API layer, which was not my goal.

src/test/java/io/supertokens/test/TestingProcessManager.java
```diff
@@ -145,4 +150,23 @@ EventAndException checkOrWaitForEvent(PROCESS_STATE state, long timeToWaitMS)
             return e;
         }
     }
+
+    /**
+     * Utility function to wrap tests with, as they require TestingProcess
+     */
+    public static void withProcess(ProcessConsumer consumer) throws Exception {
```
rishabhpoddar, line 157: - Maybe you should add an optional param (via overloading) which will allow for not MySQL based dbs (cause session tests can work with mongodb as well).
- This would also need to go in each of the plugins, as they have their own `TestingProcessManager` class.

src/main/java/io/supertokens/webserver/api/thirdparty/getUsersByEmail/GetUsersByEmailAPI.java
rishabhpoddar: Yes. But as discussed, please follow this coding convention that we have.

src/test/java/io/supertokens/test/EmailValidatorTest.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.List;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class EmailValidatorTest {
+
+    @Test
+    public void testShouldReturnTrueForValidEmails() {
+        List<String> validEmails = Arrays.asList("john.doe@example.com", "jd@e.co");
```
rishabhpoddar, line 32: Also add emails like `test+key@example.com`

src/main/java/io/supertokens/thirdparty/ThirdParty.java
```diff
@@ -158,6 +162,18 @@ public static UserPaginationContainer getUsers(Main main,
...
     }
 
+    public static UserInfo[] getUsersByEmail(Main main, String email) throws StorageQueryException, InvalidInputException {
+        if (email == null) {
+            throw new InvalidInputException("email cannot be empty");
+        }
+
+        if (!EmailValidator.isValid(email)) {
+            throw new InvalidInputException("email is invalid");
+        }
```
rishabhpoddar, line 172: Validation should go in API.

rishabhpoddar (changes requested):

src/test/java/io/supertokens/test/GetUsersByEmailAPITest2_8.java
rishabhpoddar: By raw function, I meant The function that the API uses (`ThirdParty. getUsersByEmail`).. sorry, should have been more clear about this in my prev comment.

rishabhpoddar: But you don't need to do this i guess cause you already implemented these tests via the API itself.

src/test/java/io/supertokens/test/GetUsersByEmailAPITest2_8.java
```diff
@@ -0,0 +1,201 @@
...
+            JsonObject response = io.supertokens.test.httpRequest.HttpRequest
+                    .sendGETRequest(process.getProcess(), "",
+                            "http://localhost:3567/recipe/users/by-email", query, 1000,
+                            1000,
+                            null, Utils.getCdiVersion2_8ForTests(), RECIPE_ID.THIRD_PARTY.toString());
+
+            // then
+            JsonArray jsonUsers = response.get("users").getAsJsonArray();
+
+            jsonUsers.forEach(jsonUser -> assertEquals("john.doe@example.com", jsonUser.getAsJsonObject().get("email").getAsString()));
```
rishabhpoddar, line 81: You may want to check that their userId is different and as per what you used when signing up

rishabhpoddar (changes requested):

src/main/java/io/supertokens/webserver/api/thirdparty/getUsersByEmail/GetUsersByEmailAPI.java
```diff
@@ -46,7 +47,7 @@ public String getPath() {
     @Override
     protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOException, ServletException {
         try {
-            String email = req.getParameter("email");
+            String email = InputParser.getQueryParamOrThrowError(req, "email", false);
 
             UserInfo[] users = ThirdParty.getUsersByEmail(super.main, email);
```
rishabhpoddar, line 52: Doesn't this function call complain about `email` being null.. Cause in other APIs, I have done `assert email != null;`. Like in the `SignInAPI` for emailpassword.

src/main/java/io/supertokens/utils/EmailValidator.java
rishabhpoddar: I would argue that having a class for just email validation is a bit too much.. a class called `Validators` which contains logic for email validation and other types of validation makes sense though.

The reason I think it's a bit too much is cause over time, we may have tons of validator classes, each containing just one function... But OK - I can accept this, as we really need to move faster.

src/main/java/io/supertokens/webserver/api/thirdparty/getUsersByEmail/GetUsersByEmailAPI.java
```diff
@@ -0,0 +1,67 @@
+/*
```
rishabhpoddar, line 1: Why is this file inside `api/thirdparty/getUsersByEmail/..` as opposed to being in `api/thirdparty/..`.

Please follow how other APIs are made. It just adds a lot of confusion.

rishabhpoddar (changes requested):

src/main/java/io/supertokens/webserver/api/thirdparty/getUsersByEmail/GetUsersByEmailAPI.java
dulowski-marek: It doesn't; neither in IDE nor when building

src/test/java/io/supertokens/test/TestingProcessManager.java
dulowski-marek: I moved that to a test, since that was common, but not related to process itself (although it depends on a process). Maybe we can just supply some common static function to check that (or base `SupertokensTest` class that would take care of such repetitive tasks).

rishabhpoddar (approved):

rishabhpoddar:
- Please edit the changelog.md file in the supertokens-core
- You need to make changes to the mysql & postgresql plugins as well to have their queries.

rishabhpoddar (approved):

rishabhpoddar:
One point we didn't discuss is that if a user signs up with two social providers with the same email, and we use this API with that email, we will get that user's info twice - but the `thirdPartyInfo` in that will be different in each of those objects.

I think this is OK. What do you think?

dulowski-marek:
> One point we didn't discuss is that if a user signs up with two social providers with the same email, and we use this API with that email, we will get that user's info twice - but the `thirdPartyInfo` in that will be different in each of those objects.
> 
> I think this is OK. What do you think?

For me this will be one of the primary usecases of this API. We could possibly map the result to a new object with user data and array of `thirdPartyData`, but I think this is redundant. The current API is ok for me.

rishabhpoddar:
As discussed @dulowski-marek , please remove email validation logic entirely from the core - more explantion [here](https://github.com/supertokens/supertokens-core/wiki/Recipe-development-guideline-for-the-backend-and-frontend-recipes) - Section 3b) (iv)

dulowski-marek:
@rishabhpoddar Can this be merged now, that I removed the validation?

src/main/java/io/supertokens/thirdparty/ThirdParty.java
```diff
@@ -164,8 +161,8 @@ public static UserPaginationContainer getUsers(Main main,
     }
 
     public static UserInfo[] getUsersByEmail(Main main, @Nonnull String email) throws StorageQueryException, InvalidInputException {
-        if (!EmailValidator.isValid(email)) {
-            throw new InvalidInputException("email is invalid");
+        if (email.equals("")) {
```
rishabhpoddar, line 164: We should not even do this check. No validation of user input means no validation.

rishabhpoddar (changes requested):

rishabhpoddar (approved):

Issue #275, opened 2021-07-12
Update user's email and password in EmailPassword


dulowski-marek:
### Goal
User should be able to update their email and password.

### Problems and assumptions
1. **Session validation** - What should happen with existing sessions if the user changes their password? It seems that we should revoke all the existing sessions, as they were established using an old grant (old password).
2. **API structure** - Should this be one request (endpoint)? I think that changing the email without changing password is a valid usecase. This makes me think that those should be separate endpoints, or at least separate routine (business logic).
    - What should happen with existing sessions if user changes their email? Should they be invalidated as well?
3. **Permissions** - Is changing the user's email/password by somebody else (e.g. admin) on our roadmap? If so, we should structure the API in a way that separates identifying a user the credentials are changed for from determining permissions for actor requesting the change.

The same problem as in [Delete User API](https://github.com/supertokens/supertokens-core/issues/273) is present as well. How are the backend SDK requests authorized in core?
4. **Second factor authorization** - Are there scenarios in which we should send a verification link (email, sms, notification) to authorize the change? Should this be done by core or backend SDK?

PR #291, opened 2021-08-01
275: Update user's password

dulowski-marek:
## Summary of change
This exposes an endpoint: `PUT /recipe/user/credentials` accepting the following body:
```ts
{
    "userId": string, // required
    "email": string, 
    "password": string
}
```

## Related issues
- https://github.com/supertokens/supertokens-core/issues/275

## Test Plan
tbd

## Documentation changes
- CDI changes

- [ ] ~`coreDriverInterfaceSupported.json` file has been updated (if needed)~
- [ ] ~Changes to the version if needed~
   - ~In `build.gradle`~
- [ ] ~If there are new dependencies that have been added in `build.gradle`, please make sure to add them in `implementationDependencies.json`.~

- [x] Swagger CDI spec
- [x] API tests
- [x] Logic tests


rishabhpoddar:
- Please start by making changes to the CDI spec.
- This API is supposed to allow changing of a user's password or email by simply editing those fields in the db. No need to create any additional token etc.. So an APi like:
   ```
   /recipe/user PUT
   {
      userId: string,
      newPassword?: string,
      email?: string
   }
   ```
   "?" means they are optional. If the user has provided both those fields, both of them are changed. And the response can be:
   ```
   {
      status: "OK" | "EMAIL_ALREADY_EXISTS_ERROR"
   }
   ```
   The password validator would need to run in the backend SDK inside the recipe function that is used to call this API.
- Due to this large difference between what was expected and what is implemented, you should close this PR and start this from scratch.

dulowski-marek:
Why do you think such API would be better than 2 endpoints allowing to separately change user's credentials? How is that different from password reset?
How do you want to approach change password request validation, if you can change the email in the same request? 

There are reasons why I chose to do two separate requests and token generation for password change. I don't know yours for the API design that you proposed at this moment, so please explain.

rishabhpoddar:
> Why do you think such API would be better than 2 endpoints allowing to separately change user's credentials?

Because this is not a password reset flow.. it's simply one of the CRUD operations we provide to the backend against the objects of this recipe.


> How do you want to approach change password request validation, if you can change the email in the same request?

What is password request validation?

> There are reasons why I chose to do two separate requests and token generation for password change. I don't know yours for the API design that you proposed at this moment, so please explain.

My logic is simply to provide CRUD operation on the objects of this recipe. That's all. The actual password reset flow is different. Happy to discuss on a call.

dulowski-marek:
Then this PR provides simple CRUD without session removal logic.

CHANGELOG.md
```diff
@@ -16,7 +16,8 @@ https://github.com/supertokens/supertokens-core/issues/255
 ### Added
 - Add `GET /recipe/users/by-email?email=john@example.com` endpoint for ThirdParty recipe to fetch all users with given
  email
-
+- Add dedicated `POST /recipe/user/password/change/token` and `POST /recipe/user/password/change` endpoints allowing to
```
rishabhpoddar, line 19: The changelog message is wrong.

src/main/java/io/supertokens/webserver/api/emailpassword/UpdateEmailOrPasswordAPI.java
```diff
@@ -0,0 +1,67 @@
...
+import java.io.IOException;
+
+public class UpdateEmailOrPasswordAPI extends WebserverAPI {
+    public UpdateEmailOrPasswordAPI(Main main) {
+        super(main, RECIPE_ID.EMAIL_PASSWORD.toString());
+    }
+
+    @Override
+    public String getPath() {
+        return "/recipe/user/credentials";
```
rishabhpoddar, line 40: Should change to `/recipe/user`

src/main/java/io/supertokens/webserver/api/emailpassword/UpdateEmailOrPasswordAPI.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Override
+    protected void doPut(HttpServletRequest req, HttpServletResponse resp) throws IOException, ServletException {
+        JsonObject input = InputParser.parseJsonObjectOrThrowError(req);
+        String userId = InputParser.parseStringOrThrowError(input, "method", false);
```
rishabhpoddar, line 46: - Mistake - no "method", should be "userId"?
- Not tested API properly.

src/test/java/io/supertokens/test/emailpassword/UpdateUsersEmailAndPasswordTest.java
```diff
@@ -0,0 +1,112 @@
+/*
```
rishabhpoddar, line 1: - Need to add API test as well
- Add a test to update info of a user if they don't exist - should return status: "UNKNOWN_USER_ID"

src/main/java/io/supertokens/webserver/api/emailpassword/UpdateEmailOrPasswordAPI.java
```diff
@@ -0,0 +1,67 @@
...
+        if (email == null && password == null) {
+            throw new ServletException(new BadRequestException("You have to provide either email or password."));
+        }
+
+        try {
+            EmailPassword.updateUsersEmailOrPassword(main, userId, email, password);
+
+            JsonObject result = new JsonObject();
+            result.addProperty("status", "OK");
+            super.sendJsonResponse(200, result, resp);
```
rishabhpoddar, line 61: The API output should be 
```
{
   status: "OK" | "UNKNOWN_USER_ID_ERROR"
}
```

rishabhpoddar (changes requested):

src/main/java/io/supertokens/emailpassword/EmailPassword.java
```diff
@@ -188,6 +188,27 @@ public static void resetPassword(Main main, String token, String password)
         }
     }
 
+    public static void updateUsersEmailOrPassword(Main main, @Nonnull String userId, @Nullable String email, @Nullable String password)
+            throws StorageQueryException, StorageTransactionLogicException {
+        EmailPasswordSQLStorage storage = StorageLayer.getEmailPasswordStorage(main);
+
+        storage.startTransaction(transaction -> {
+            if (email != null) {
```
rishabhpoddar, line 196: - You will first want to take a lock on all this userId's row by using a select for update query (which should already exist in the plugin interface).
- Then if the userId doesn't exist, you would return `"UNKNOWN_USER_ID_ERROR"`, else continue.

rishabhpoddar (changes requested):

rishabhpoddar (approved):

PR #302, opened 2021-08-26
Add the ability to create JWTs using supertokens

nkshah2:
## Summary of change
This PR adds a new recipe for creating JWTs using supertokens.
- Adds APIs for creating signed JWTs and for fetching JWKs for verification
- Adds auth0/java-jwt which is used to create JWTs (and verify in unit tests)
- Creates a table for storing keys against signing algorithms

## Related issues
- https://github.com/supertokens/for-zenhub/issues/16

## Related Pull Requests
- https://github.com/supertokens/supertokens-mongodb-plugin/pull/8

## Test Plan
- Unit tests added for JWT creating and JWKs (including verifying using the returned keys)
- API tests added for `/recipe/jwt POST` and `/recipe/jwt/jwks GET`

## Documentation changes
These changes will be made after functionality is implemented in the backend SDK
- [ ] Db schema change
- [ ] API gateway / Hasura integration docs
- [ ] Changing getJWTPayload to getAccessTokenPayload and other similar functions
- [ ] Change the use of the word JWT to access token in the docs (in most places)

   - ~~In `build.gradle`~~

- [x] Unit tests for JWKs
- [x] API tests
- [x] Update CHANGELOG
- [x] Add test that tests race condition with duplicate key id and simultaneous writes for nosql plugins
- [x] Add comments to existing isolation tests for storage
- [x] Add logic to create one key for each supported algorithm type on startup (key should only be created if one does not already exist in storage)


src/main/java/io/supertokens/jwt/JWTSigningFunctions.java
```diff
@@ -0,0 +1,124 @@
...
+        Algorithm signingAlgorithm = getAlgorithmFromString(main, algorithm, keyToUse);
+
+        // Create the claims for the JWT header
+        Map<String, Object> headerClaims = new HashMap<>();
+        headerClaims.put("alg", algorithm.toUpperCase()); // All examples in the RFC have the algorithm in upper case
+        headerClaims.put("typ", "JWT");
+        headerClaims.put("kid", keyToUse.keyId);
+
+        long currentTimeInMillis = System.currentTimeMillis();
+        long jwtExpiry = (currentTimeInMillis / 1000) + (jwtValidity); // JWT Expiry is seconds from epoch not millis
```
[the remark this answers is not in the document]
nkshah2: Since JWTs seem to use seconds for expiry, issued time, not before etc I think it makes sense to use seconds

src/main/java/io/supertokens/webserver/api/jwt/JWTSigningAPI.java
```diff
@@ -0,0 +1,76 @@
...
+            return;
+        }
+
+        String jwksDomain = InputParser.parseStringOrThrowError(input, "jwksDomain", false);
+        assert jwksDomain != null;
+
+        JsonObject payload = InputParser.parseJsonObjectOrThrowError(input, "payload", false);
+        assert payload != null;
+
+        long validity = InputParser.parseLongOrThrowError(input, "validity", false);
```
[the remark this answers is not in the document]
nkshah2: We can do this but using 0 as validity does not cause any issues. Yes the JWT would not be useable at all if it is 0 but since it isn't a problem should the core validate and throw an error in this case? Its a small change so I don't mind doing it but it doesn't feel necessary.

src/main/java/io/supertokens/webserver/api/jwt/JWKAPI.java
```diff
@@ -0,0 +1,62 @@
+/*
```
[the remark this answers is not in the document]
nkshah2: So this was probably my OCD kicking in but JWKs is just plural, since the API is to fetch a JWK (even though its an array only one is used) for validation `JWKAPI` seemed like a better name. Im agnostic though so I dont mind either.

src/main/java/io/supertokens/webserver/api/jwt/JWKAPI.java
```diff
@@ -0,0 +1,62 @@
...
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOException, ServletException {
+        try {
+            List<JsonObject> jwks = JWTSigningFunctions.getJWKS(main);
+            JsonObject reply = new JsonObject();
+            JsonArray jwksJsonArray = new JsonParser().parse(new Gson().toJson(jwks)).getAsJsonArray();
+            reply.add("keys", jwksJsonArray);
+            super.sendJsonResponse(200, reply, resp);
+        } catch (StorageQueryException | StorageTransactionLogicException | NoSuchAlgorithmException
+                | InvalidKeySpecException e) {
```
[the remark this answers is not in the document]
nkshah2: Both exceptions are thrown when creating a public key from its string representation. For example: 

```java
KeyFactory kf = KeyFactory.getInstance("RSA");
 return (RSAPublicKey) kf.generatePublic(keySpec);
```

would throw `NoSuchAlgorithmException` if the string passed to `KeyFactory.getInstance` is an algorithm that JAVAs cryptography doesnt support (or if it some garbage string for example), and would throw `InvalidKeySpecException` "if the given key specification is inappropriate for this key factory to produce a public key" (picked up from the javadocs for the function)

src/main/java/io/supertokens/Main.java
```diff
@@ -175,6 +176,7 @@ private void init() throws IOException {
         // init signing keys
         AccessTokenSigningKey.init(this);
         RefreshTokenKey.init(this);
+        JWTSigningKey.init(this);
```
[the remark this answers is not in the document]
nkshah2: Done

src/main/java/io/supertokens/jwt/JWTSigningKey.java
```diff
@@ -0,0 +1,182 @@
...
+    private final Main main;
+
+    // Used to validate the algorithm sent from the backend SDK and to fetch an appropriate key when creating JWTs
+    public static final Set<String> supportedAlgorithms = Set.of("rs256");
+
+    private JWTSigningKey(Main main) {
+        this.main = main;
+    }
+
+    public static void init(Main main) {
```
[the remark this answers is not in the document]
nkshah2: Done

src/main/java/io/supertokens/jwt/JWTSigningKey.java
```diff
@@ -0,0 +1,182 @@
...
+
+    public static JWTSigningKey getInstance(Main main) {
+        JWTSigningKey instance = (JWTSigningKey) main.getResourceDistributor().getResource(RESOURCE_KEY);
+        if (instance == null) {
+            init(main);
+        }
+        return (JWTSigningKey) main.getResourceDistributor().getResource(RESOURCE_KEY);
+    }
+
+    public synchronized List<JWTSigningKeyInfo> getAllSigningKeys()
```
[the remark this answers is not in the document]
nkshah2: It doesnt need to be, done

src/main/java/io/supertokens/jwt/JWTSigningKey.java
```diff
@@ -0,0 +1,182 @@
...
+                keys = keysFromStorage;
+            }
+
+            return keys;
+        }
+
+        throw new QuitProgramException("Unsupported storage type detected");
+    }
+
+    public synchronized JWTSigningKeyInfo getKeyForAlgorithm(String algorithm)
```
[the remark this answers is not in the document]
nkshah2: It doesnt need to be, done

src/main/java/io/supertokens/jwt/JWTSigningKey.java
```diff
@@ -0,0 +1,182 @@
...
+                keyInfo = keyFromStorage;
+            }
+
+            return keyInfo;
+        }
+
+        throw new QuitProgramException("Unsupported storage type detected");
+    }
+
+    private String getAlgorithmType(String algorithm) throws NoSuchAlgorithmException {
```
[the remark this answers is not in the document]
nkshah2: Algorithm handling has been moved to an enum `JWTSigningKey.SupportedAlgorithms`, the enum now exposes a way to get algorithm strings and types instead of having functions like this one.

src/main/java/io/supertokens/jwt/JWTSigningKey.java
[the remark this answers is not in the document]
nkshah2: Done

src/main/java/io/supertokens/jwt/JWTSigningKey.java
```diff
@@ -0,0 +1,182 @@
...
+               JWTSigningKeyInfo keyInfo;
+               String algorithmType;
+
+               try {
+                   algorithmType = getAlgorithmType(algorithm);
+               } catch (NoSuchAlgorithmException e) {
+                   throw new StorageTransactionLogicException(e);
+               }
+
+               JWTSigningKeyInfo keyFromStorage = sqlStorage.getLatestJWTSigningKeyForAlgorithm_Transaction(con, algorithmType);
```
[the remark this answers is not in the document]
nkshah2: Done

src/main/java/io/supertokens/jwt/JWTSigningKey.java
```diff
@@ -0,0 +1,182 @@
...
+            } catch (NoSuchAlgorithmException e) {
+                throw new StorageTransactionLogicException(e);
+            }
+
+            JWTSigningKeyInfo keyFromStorage = noSQLStorage.getLatestJWTSigningKeyForAlgorithm_Transaction(algorithmType);
+
+            if (keyFromStorage == null) {
+                long currentTimeInMillis = System.currentTimeMillis();
+                Utils.PubPriKey newKey = Utils.generateNewPubPriKey();
+                keyInfo = new JWTSigningKeyInfo(Utils.getUUID(), newKey.publicKey, newKey.privateKey, currentTimeInMillis, algorithm, algorithmType);
```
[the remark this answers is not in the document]
nkshah2: Done

src/main/java/io/supertokens/inmemorydb/Start.java
```diff
@@ -58,6 +60,7 @@
     private static final String APP_ID_KEY_NAME = "app_id";
     private static final String ACCESS_TOKEN_SIGNING_KEY_NAME = "access_token_signing_key";
     private static final String REFRESH_TOKEN_KEY_NAME = "refresh_token_key";
+    private static final String JWT_SIGNING_KEY_NAME = "jwt_signing_token_key";
```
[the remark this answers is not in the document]
nkshah2: Done

src/main/java/io/supertokens/inmemorydb/queries/GeneralQueries.java
```diff
@@ -392,4 +401,89 @@ public static long getUsersCount(Start start, RECIPE_ID[] includeRecipeIds) thro
             this.recipeId = recipeId;
         }
     }
+
+    public static List<JWTSigningKeyInfo> getJWTSigningKeys_Transaction(Start start, Connection con)
```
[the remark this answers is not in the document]
nkshah2: Done

src/main/java/io/supertokens/inmemorydb/queries/GeneralQueries.java
```diff
@@ -392,4 +401,89 @@ public static long getUsersCount(Start start, RECIPE_ID[] includeRecipeIds) thro
...
+
+            while(result.next()) {
+                keys.add(JWTSigningKeyInfoRowMapper.getInstance().mapOrThrow(result));
+            }
+
+            return keys;
+        }
+    }
+
+    private static class JWTSigningKeyInfoRowMapper implements RowMapper<JWTSigningKeyInfo, ResultSet> {
```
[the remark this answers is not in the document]
nkshah2: Done

src/main/java/io/supertokens/inmemorydb/queries/GeneralQueries.java
```diff
@@ -392,4 +401,89 @@ public static long getUsersCount(Start start, RECIPE_ID[] includeRecipeIds) thro
...
+            return INSTANCE;
+        }
+
+        @Override
+        public JWTSigningKeyInfo map(ResultSet result) throws Exception {
+            return new JWTSigningKeyInfo(result.getString("key_id"), result.getString("public_key"), result.getString("private_key"), result.getLong("created_at"), result.getString("algorithm"), result.getString("algorithm_type"));
+        }
+    }
+
+    public static void setJWTSigningKeyInfo_Transaction(Start start, Connection con, JWTSigningKeyInfo info)
```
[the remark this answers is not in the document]
nkshah2: Done

src/main/java/io/supertokens/inmemorydb/queries/GeneralQueries.java
```diff
@@ -392,4 +401,89 @@ public static long getUsersCount(Start start, RECIPE_ID[] includeRecipeIds) thro
...
+            pst.setString(2, info.publicKey);
+            pst.setString(3, info.privateKey);
+            pst.setLong(4, info.createdAtTime);
+            pst.setString(5, info.algorithm);
+            pst.setString(6, info.algorithmType);
+            pst.executeUpdate();
+        }
+    }
+
+    public static JWTSigningKeyInfo getLatestJWTSigningKey_Transaction(Start start, Connection con)
```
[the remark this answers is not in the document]
nkshah2: Removed the function

src/main/java/io/supertokens/inmemorydb/queries/GeneralQueries.java
```diff
@@ -392,4 +401,89 @@ public static long getUsersCount(Start start, RECIPE_ID[] includeRecipeIds) thro
...
+
+            if (result.next()) {
+                return JWTSigningKeyInfoRowMapper.getInstance().mapOrThrow(result);
+            }
+        }
+
+        return null;
+    }
+
+    public static JWTSigningKeyInfo getLatestJWTSigningKeyForAlgorithm_Transaction(Start start, Connection con, String algorithm)
```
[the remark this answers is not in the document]
nkshah2: Removed in favour of `getJWTSigningKeys_Transaction` as discussed

src/main/java/io/supertokens/inmemorydb/Start.java
```diff
@@ -796,4 +799,37 @@ public long getThirdPartyUsersCount() throws StorageQueryException {
...
+        Connection sqlCon = (Connection) con.getConnection();
+        try {
+            GeneralQueries.setJWTSigningKeyInfo_Transaction(this, sqlCon, info);
+        } catch (SQLException e) {
+            throw new StorageQueryException(e);
+        }
+    }
+
+    @Override
+    public JWTSigningKeyInfo getLatestJWTSigningKeyForAlgorithm_Transaction(TransactionConnection con, String algorithm)
```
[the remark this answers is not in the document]
nkshah2: Done

src/main/java/io/supertokens/inmemorydb/queries/GeneralQueries.java
```diff
@@ -392,4 +401,89 @@ public static long getUsersCount(Start start, RECIPE_ID[] includeRecipeIds) thro
...
+
+    public static void setJWTSigningKeyInfo_Transaction(Start start, Connection con, JWTSigningKeyInfo info)
+            throws SQLException {
+
+        String QUERY = "INSERT INTO " + Config.getConfig(start).getJWTSigningKeysTable()
+                + "(key_id, public_key, private_key, created_at, algorithm, algorithm_type) VALUES(?, ?, ?, ?, ?, ?)";
+
+        try (PreparedStatement pst = con.prepareStatement(QUERY)) {
+            pst.setString(1, info.keyId);
+            pst.setString(2, info.publicKey);
```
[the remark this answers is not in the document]
nkshah2: Done. Introduced two new classes (in the plugin interface) for Symmetric and Asymmetric key handling

src/main/java/io/supertokens/inmemorydb/queries/GeneralQueries.java
```diff
@@ -392,4 +401,89 @@ public static long getUsersCount(Start start, RECIPE_ID[] includeRecipeIds) thro
...
+            pst.setString(3, info.privateKey);
+            pst.setLong(4, info.createdAtTime);
+            pst.setString(5, info.algorithm);
+            pst.setString(6, info.algorithmType);
```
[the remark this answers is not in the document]
nkshah2: Done, removed algorithmType.

src/main/java/io/supertokens/inmemorydb/queries/JWTSigningQueries.java
```diff
@@ -0,0 +1,35 @@
...
+public class JWTSigningQueries {
+    static String getQueryToCreateJWTSigningTable(Start start) {
+        return "CREATE TABLE IF NOT EXISTS " + Config.getConfig(start).getJWTSigningKeysTable() + " ("
+                + "key_id VARCHAR(255) NOT NULL," + "public_key TEXT NOT NULL,"
+                + "private_key TEXT NOT NULL," + "algorithm VARCHAR(10) NOT NULL,"
+                + "algorithm_type VARCHAR(10) NOT NULL," + "created_at BIGINT UNSIGNED,"
+                + "PRIMARY KEY(key_id));";
+    }
+
+    static String getQueryToFetchLatestKey(Start start) {
```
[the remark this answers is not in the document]
nkshah2: Removed

src/main/java/io/supertokens/jwt/JWTSigningFunctions.java
```diff
@@ -0,0 +1,124 @@
...
+            jwk.addProperty("alg", currentKeyInfo.algorithm.toUpperCase());
+            jwk.addProperty("use", "sig");
+
+            jwks.add(jwk);
+        }
+
+        return jwks;
+    }
+
+    private static JWTSigningKeyInfo getKeyToUse(Main main, String algorithm)
```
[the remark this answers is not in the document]
nkshah2: Done, removed this function and used the static functon in JWTSigningKey directly

src/main/java/io/supertokens/jwt/JWTSigningFunctions.java
```diff
@@ -0,0 +1,124 @@
...
+        long currentTimeInMillis = System.currentTimeMillis();
+        long jwtExpiry = (currentTimeInMillis / 1000) + (jwtValidity); // JWT Expiry is seconds from epoch not millis
+
+        // Add relevant claims to the payload, note we only add/override ones that we absolutely need to.
+        Map<String, Object> jwtPayload = new Gson().fromJson(payload, HashMap.class);
+        jwtPayload.put("iss", jwksDomain);
+        jwtPayload.put("exp", jwtExpiry);
+        jwtPayload.put("iat", currentTimeInMillis / 1000); // JWT uses seconds from epoch not millis
+
+        // TODO: Add custom payload
```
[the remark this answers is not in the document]
nkshah2: Done

src/main/java/io/supertokens/jwt/JWTSigningKey.java
```diff
@@ -0,0 +1,182 @@
...
+                   } catch (NoSuchAlgorithmException e) {
+                       throw new StorageTransactionLogicException(e);
+                   }
+               } else {
+                   keyInfo = keyFromStorage;
+               }
+
+               sqlStorage.commitTransaction(con);
+               return keyInfo;
+            });
```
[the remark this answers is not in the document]
nkshah2: Done, since NoSuchAlgorithm is an exception that Java's cryptography throws I am now using a custom UnsupportedAlgorithmException for our use case (the names do seem confusing, maybe I should call it something else?)

src/main/java/io/supertokens/webserver/api/jwt/JWTSigningAPI.java
```diff
@@ -0,0 +1,76 @@
...
+        return "/recipe/jwt";
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws IOException, ServletException {
+        JsonObject input = InputParser.parseJsonObjectOrThrowError(req);
+        String algorithm = InputParser.parseStringOrThrowError(input, "algorithm", false);
+        assert algorithm != null;
+
+        if (!JWTSigningKey.isJWTAlgorithmSupported(algorithm)) {
```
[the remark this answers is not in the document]
nkshah2: Done, the validation is now handled when converting the string value to the `JWTSigningKey.SupportAlgorithms` enum when createJWT is called.

src/main/java/io/supertokens/jwt/JWTSigningKey.java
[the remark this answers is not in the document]
nkshah2: The exception is now called `UnsupportedJWTSigningAlgorithmException`

src/main/java/io/supertokens/jwt/JWTSigningKey.java
```diff
@@ -0,0 +1,182 @@
...
+               }
+
+               JWTSigningKeyInfo keyFromStorage = sqlStorage.getLatestJWTSigningKeyForAlgorithm_Transaction(con, algorithmType);
+
+               if (keyFromStorage == null) {
+                   try {
+                       long currentTimeInMillis = System.currentTimeMillis();
+                       Utils.PubPriKey newKey = Utils.generateNewPubPriKey();
+                       keyInfo = new JWTSigningKeyInfo(Utils.getUUID(), newKey.publicKey, newKey.privateKey, currentTimeInMillis, algorithm, algorithmType);
+                       sqlStorage.setJWTSigningKey_Transaction(con, keyInfo);
```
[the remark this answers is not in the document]
nkshah2: Done

src/main/java/io/supertokens/inmemorydb/queries/GeneralQueries.java
```diff
@@ -392,4 +401,89 @@ public static long getUsersCount(Start start, RECIPE_ID[] includeRecipeIds) thro
             this.recipeId = recipeId;
         }
     }
+
+    public static List<JWTSigningKeyInfo> getJWTSigningKeys_Transaction(Start start, Connection con)
+            throws SQLException, StorageQueryException {
+
+        // TODO: does this need locking?
```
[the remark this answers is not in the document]
nkshah2: Done for in memory DB

src/main/java/io/supertokens/webserver/api/jwt/JWTSigningAPI.java
```diff
@@ -0,0 +1,76 @@
...
+
+        long validity = InputParser.parseLongOrThrowError(input, "validity", false);
+
+        try {
+            String jwt = JWTSigningFunctions.createJWTToken(main, algorithm, payload, jwksDomain, validity);
+            JsonObject reply = new JsonObject();
+            reply.addProperty("status", "OK");
+            reply.addProperty("jwt", jwt);
+            super.sendJsonResponse(200, reply, resp);
+        } catch (Exception e) {
```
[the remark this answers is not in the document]
nkshah2: The error is no longer needed, changes to the spec are made in this PR: https://github.com/supertokens/core-driver-interface/pull/17

src/main/java/io/supertokens/jwt/JWTSigningKey.java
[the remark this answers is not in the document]
nkshah2: Done

src/main/java/io/supertokens/inmemorydb/Start.java
```diff
@@ -796,4 +799,37 @@ public long getThirdPartyUsersCount() throws StorageQueryException {
...
+        }
+    }
+
+    @Override
+    public void setJWTSigningKey_Transaction(TransactionConnection con, JWTSigningKeyInfo info)
+            throws StorageQueryException {
+        Connection sqlCon = (Connection) con.getConnection();
+        try {
+            GeneralQueries.setJWTSigningKeyInfo_Transaction(this, sqlCon, info);
+        } catch (SQLException e) {
```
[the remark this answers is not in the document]
nkshah2: Done

src/main/java/io/supertokens/webserver/api/jwt/JWKAPI.java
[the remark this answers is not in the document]
nkshah2: Done

src/main/java/io/supertokens/webserver/api/jwt/JWTSigningAPI.java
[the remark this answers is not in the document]
rishabhpoddar: How about checking if it is >= 0? We don't want to allow negative values?

src/main/java/io/supertokens/inmemorydb/queries/JWTSigningQueries.java
```diff
@@ -16,20 +16,93 @@
...
+        public JWTSigningKeyInfo map(ResultSet result) throws Exception {
+            String keyId = result.getString("key_id");
+            String keyString = result.getString("key_string");
+            long createdAt = result.getLong("created_at");
+            String algorithm = result.getString("algorithm");
+
+            if (keyString.contains("|")) {
+                return JWTAsymmetricSigningKeyInfo.withKeyString(keyId, createdAt, algorithm, keyString);
+            } else {
+                return new JWTSymmetricSigningKeyInfo(keyId, createdAt, algorithm, keyString);
```
rishabhpoddar, line 87: Why is one using `withKeyString` and the other just calling the constructor?

src/main/java/io/supertokens/inmemorydb/queries/JWTSigningQueries.java
```diff
@@ -16,20 +16,93 @@
...
+
+        try (PreparedStatement pst = con.prepareStatement(QUERY)) {
+            pst.setString(1, info.keyId);
+            pst.setString(2, info.keyString);
+            pst.setLong(3, info.createdAtTime);
+            pst.setString(4, info.algorithm);
+            pst.executeUpdate();
+        } catch (SQLException e) {
+            throw new StorageTransactionLogicException(e);
+        }
```
rishabhpoddar, line 106: Why are you not throwing the `SQLException` directly?

src/main/java/io/supertokens/jwt/JWTSigningKey.java
```diff
@@ -0,0 +1,247 @@
...
+        SessionStorage storage = StorageLayer.getSessionStorage(main);
+
+        if (storage.getType() == STORAGE_TYPE.SQL) {
+            SessionSQLStorage sqlStorage = (SessionSQLStorage) storage;
+
+            return sqlStorage.startTransaction(con -> {
+               List<JWTSigningKeyInfo> keys;
+               List<JWTSigningKeyInfo> keysFromStorage = sqlStorage.getJWTSigningKeys_Transaction(con);
+
+               if (keysFromStorage == null) {
```
rishabhpoddar, line 88: When would `keysFromStorage` be `null`? If there are no keys, shouldn't the function return an empty `List`?

src/main/java/io/supertokens/jwt/JWTSigningKey.java
```diff
@@ -0,0 +1,247 @@
...
+                    JWTSigningKeyInfo keyInfo = null;
+
+                    List<JWTSigningKeyInfo> keysFromStorage = sqlStorage.getJWTSigningKeys_Transaction(con);
+
+                    // If there are no keys in storage, generate a new one
+                    if (keysFromStorage.isEmpty()) {
+                        try {
+                            keyInfo = generateKeyForAlgorithm(algorithm);
+                            sqlStorage.setJWTSigningKey_Transaction(con, keyInfo);
+                        } catch (NoSuchAlgorithmException | DuplicateKeyIdException | UnsupportedJWTSigningAlgorithmException e) {
```
rishabhpoddar, line 142: If `DuplicateKeyIdException` is thrown, you need to retry this operation here.

src/main/java/io/supertokens/jwt/JWTSigningKey.java
```diff
@@ -0,0 +1,247 @@
...
+        SessionStorage storage = StorageLayer.getSessionStorage(main);
+
+        if (storage.getType() == STORAGE_TYPE.SQL) {
+            SessionSQLStorage sqlStorage = (SessionSQLStorage) storage;
+
+            try {
+                return sqlStorage.startTransaction(con -> {
+                    JWTSigningKeyInfo keyInfo = null;
+
+                    List<JWTSigningKeyInfo> keysFromStorage = sqlStorage.getJWTSigningKeys_Transaction(con);
```
rishabhpoddar, line 135: Have you tried to run a SELECT FOR UPDATE query on the whole table in mysql and postgresql and seen if there is locking being done?

src/main/java/io/supertokens/jwt/JWTSigningKey.java
```diff
@@ -0,0 +1,247 @@
...
+                            keyInfo = generateKeyForAlgorithm(algorithm);
+                            sqlStorage.setJWTSigningKey_Transaction(con, keyInfo);
+                        } catch (NoSuchAlgorithmException | DuplicateKeyIdException | UnsupportedJWTSigningAlgorithmException e) {
+                            throw new StorageTransactionLogicException(e);
+                        }
+                    } else {
+                        // Loop through the keys and find the first one for the algorithm
+                        for (int i = 0; i < keysFromStorage.size(); i++) {
+                            JWTSigningKeyInfo currentKey = keysFromStorage.get(i);
+                            if (currentKey.algorithm.equalsIgnoreCase(algorithm.name())) {
```
rishabhpoddar, line 149: Should you just implement a `.equals` function in the algorithm class so that you can do `currentKey.algorithm.equals(algorithm)`

src/main/java/io/supertokens/jwt/JWTSigningKey.java
```diff
@@ -0,0 +1,247 @@
...
+                                keyInfo = currentKey;
+                                break;
+                            }
+                        }
+
+                        // If no key was found create a new one
+                        if (keyInfo == null) {
```
rishabhpoddar, line 156: Refactor code to prevent duplication of logic for creation of key.

src/main/java/io/supertokens/jwt/JWTSigningKey.java
```diff
@@ -0,0 +1,247 @@
...
+                                throw new StorageTransactionLogicException(e);
+                            }
+                        }
+                    }
+
+                    sqlStorage.commitTransaction(con);
+                    return keyInfo;
+                });
+            } catch (StorageTransactionLogicException e) {
+                if (e.actualException instanceof DuplicateKeyIdException) {
```
rishabhpoddar, line 170: `DuplicateKeyIdException` should not be thrown at all.

src/main/java/io/supertokens/jwt/JWTSigningKey.java
```diff
@@ -0,0 +1,247 @@
...
+            JWTSigningKeyInfo keyInfo = null;
+
+            while (true) {
+                List<JWTSigningKeyInfo> keysFromStorage = noSQLStorage.getJWTSigningKeys_Transaction();
+
+                // If there are no keys in storage, generate a new one
+                if (keysFromStorage.isEmpty()) {
+                    try {
+                        keyInfo = generateKeyForAlgorithm(algorithm);
+                        boolean success = noSQLStorage.setJWTSigningKeyInfo_Transaction(keyInfo);
```
rishabhpoddar, line 190: Wouldn't this also throw `DuplicateKeyIdException`?

src/main/java/io/supertokens/jwt/JWTSigningKey.java
```diff
@@ -0,0 +1,247 @@
...
+                            keyInfo = currentKey;
+                            break;
+                        }
+                    }
+
+                    // If no key was found create a new one
+                    if (keyInfo == null) {
+                        try {
+                            keyInfo = generateKeyForAlgorithm(algorithm);
+                            boolean success = noSQLStorage.setJWTSigningKeyInfo_Transaction(keyInfo);
```
rishabhpoddar, line 214: Refactor code to prevent duplication of logic for creation of key.

src/main/java/io/supertokens/jwt/JWTSigningKey.java
```diff
@@ -0,0 +1,247 @@
...
+        if (algorithm.getAlgorithmType().equalsIgnoreCase("rsa")) {
+            long currentTimeInMillis = System.currentTimeMillis();
+            Utils.PubPriKey newKey = Utils.generateNewPubPriKey();
+            return new JWTAsymmetricSigningKeyInfo(Utils.getUUID(), currentTimeInMillis, algorithm.name(), newKey.publicKey,
+                    newKey.privateKey);
+        }
+
+        // Ideally this should never execute because for unsupported algorithms an exception will be thrown before this method
+        // This is just so that we dont need to handle null when we call this method
+        throw new UnsupportedJWTSigningAlgorithmException();
```
rishabhpoddar, line 245: This is not good logic. If someone passes null into this function, throw some in built runtime exception and not this specific excption that we have. I believe `UnsupportedJWTSigningAlgorithmException` is thrown when creating the `SupportedAlgorithms` object?

src/main/java/io/supertokens/jwt/JWTSigningFunctions.java
```diff
@@ -0,0 +1,164 @@
...
+            throws StorageQueryException, StorageTransactionLogicException, NoSuchAlgorithmException, InvalidKeySpecException,
+            JWTCreationException, UnsupportedJWTSigningAlgorithmException, DuplicateKeyIdException {
+        // TODO: In the future we will have a way for the user to send a custom key id to use
+        JWTSigningKey.SupportedAlgorithms supportedAlgorithm;
+
+        try {
+            supportedAlgorithm = JWTSigningKey.SupportedAlgorithms.valueOf(algorithm);
+        } catch (IllegalArgumentException e) {
+            // If it enters this block then the string value provided does not match the algorithms we support
+            throw new UnsupportedJWTSigningAlgorithmException();
```
rishabhpoddar, line 66: Add a test to make sure it comes here as expected.

rishabhpoddar (changes requested):

src/main/java/io/supertokens/jwt/JWTSigningFunctions.java
```diff
@@ -56,69 +85,80 @@ public static String createJWTToken(Main main, String algorithm, JsonObject payl
...
         return jwks;
     }
 
-    private static JWTSigningKeyInfo getKeyToUse(Main main, String algorithm)
-            throws StorageQueryException, StorageTransactionLogicException, NoSuchAlgorithmException {
-        return JWTSigningKey.getInstance(main).getKeyForAlgorithm(algorithm);
-    }
-
-    private static Algorithm getAlgorithmFromString(Main main, String algorithm, JWTSigningKeyInfo keyToUse) throws NoSuchAlgorithmException, InvalidKeySpecException {
+    private static Algorithm getAuth0Algorithm(JWTSigningKey.SupportedAlgorithms algorithm, JWTSigningKeyInfo keyToUse) throws NoSuchAlgorithmException, InvalidKeySpecException,
```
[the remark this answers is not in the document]
nkshah2: Its directly using Auth0s package which calls the object "Algorithm" which I thought could be confusing since that is used for the signing algorithm as well, for the lack of a better name I used this one

src/main/java/io/supertokens/inmemorydb/queries/JWTSigningQueries.java
nkshah2: For Symmetric keys the key string is the key and does not need any extra computation, for Asymmetric keys its a helper function that splits the keystring with the "|" delimiter and then instantiates using the class constructor

src/main/java/io/supertokens/jwt/JWTSigningKey.java
nkshah2: Throwing IllegalArgument as discussed

src/main/java/io/supertokens/webserver/api/jwt/JWTSigningAPI.java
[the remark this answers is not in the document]
nkshah2: Done

src/main/java/io/supertokens/inmemorydb/queries/JWTSigningQueries.java
```diff
@@ -16,20 +16,93 @@
...
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.List;
 
 public class JWTSigningQueries {
+    private final static String lockString = "jwt_signing_keys_table_lock";
```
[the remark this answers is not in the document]
nkshah2: Done

src/main/java/io/supertokens/jwt/JWTSigningKey.java
nkshah2: So my thinking was that from the storage layer we should return a list if there are keys, or null otherwise. The storage layer should not assume that the caller does not want to handle nulls. So instead of the storage layer returning an empty List this method handles null.

src/main/java/io/supertokens/jwt/JWTSigningKey.java
nkshah2: SELECT FOR UPDATE locks the rows that are indexed (indexed is a wrong word here but for lack of better word, I mean rows that the query would traverse and determine if the row needs to be returned) from a query but doesnt lock the entire table, in theory since SELECT * would "touch" all the rows so it would work. However if the table is empty when the query runs (the first time we fetch keys, in the future if the user deletes all keys etc) then it wouldnt lock anything. Still reading up on this

src/main/java/io/supertokens/jwt/JWTSigningKey.java
nkshah2: Makes sense, but `.equals` implies matching with the same type I feel so using `.equalsString` instead

src/main/java/io/supertokens/inmemorydb/queries/JWTSigningQueries.java
nkshah2: Done

src/main/java/io/supertokens/jwt/JWTSigningFunctions.java
```diff
@@ -34,16 +37,42 @@
...
+     * @throws NoSuchAlgorithmException If there is an error when using Java's cryptography packages
+     * @throws InvalidKeySpecException If there is an error when using Java's cryptography packages
+     * @throws JWTCreationException If there is an error when creating JWTs
+     * @throws UnsupportedJWTSigningAlgorithmException If the algorithm provided does not match any of the supported algorithms
+     * @throws DuplicateKeyIdException If there is an error storing generated keys in storage
+     */
     public static String createJWTToken(Main main, String algorithm, JsonObject payload, String jwksDomain, long jwtValidity)
             throws StorageQueryException, StorageTransactionLogicException, NoSuchAlgorithmException, InvalidKeySpecException,
-            JWTCreationException {
+            JWTCreationException, UnsupportedJWTSigningAlgorithmException, DuplicateKeyIdException {
```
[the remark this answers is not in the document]
nkshah2: Done

src/main/java/io/supertokens/webserver/api/jwt/JWTSigningAPI.java
```diff
@@ -67,10 +66,14 @@ protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws I
...
             reply.addProperty("jwt", jwt);
             super.sendJsonResponse(200, reply, resp);
-        } catch (Exception e) {
+        } catch (UnsupportedJWTSigningAlgorithmException e) {
             JsonObject reply = new JsonObject();
-            reply.addProperty("status", "JWT_CREATION_ERROR");
+            reply.addProperty("status", "UNSUPPORTED_ALGORITHM_ERROR");
             super.sendJsonResponse(200, reply, resp);
+        } catch (StorageQueryException | StorageTransactionLogicException | NoSuchAlgorithmException
+                | InvalidKeySpecException | DuplicateKeyIdException e) {
```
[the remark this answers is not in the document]
nkshah2: Done

src/main/java/io/supertokens/jwt/JWTSigningKey.java
```diff
@@ -100,83 +112,136 @@ public static JWTSigningKey getInstance(Main main) {
...
+     * Used to retrieve a key for JWT validation, for a given signing algorithm. If there are no keys in storage that match the given algorithm a new one is generated.
+     * @param algorithm The signing algorithm
+     * @return {@link JWTSigningKeyInfo} key
+     * @throws UnsupportedJWTSigningAlgorithmException If there is an error in the provided algorithm when getting/generating keys
+     * @throws StorageQueryException If there is an error interacting with the database
+     * @throws StorageTransactionLogicException If there is an error interacting with the database
+     * @throws DuplicateKeyIdException If there is an error setting generated keys in storage
+     */
+    public JWTSigningKeyInfo getOrCreateAndGetKeyForAlgorithm(SupportedAlgorithms algorithm)
+            throws UnsupportedJWTSigningAlgorithmException, StorageQueryException, StorageTransactionLogicException, DuplicateKeyIdException {
```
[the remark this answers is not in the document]
nkshah2: Done

src/main/java/io/supertokens/jwt/JWTSigningKey.java
nkshah2: Done

src/main/java/io/supertokens/jwt/JWTSigningKey.java
nkshah2: Done, moved to independent functions for SQL and NOSQL separately

src/main/java/io/supertokens/jwt/JWTSigningKey.java
nkshah2: Done

src/main/java/io/supertokens/jwt/JWTSigningKey.java
nkshah2: It does, its being handled in the independent function which creates the key now. It loops if that error is thrown and retries

src/main/java/io/supertokens/jwt/JWTSigningKey.java
nkshah2: Done

src/main/java/io/supertokens/jwt/JWTSigningFunctions.java
nkshah2: Done

nkshah2:
# Tests Added
Quick overview of all the cases for which tests are added

### Unit Tests

##### JWT creation
- Testing that normal functioning of jwt creation works fine
- Testing that using an invalid algorithm throws UnsupportedJWTSigningAlgorithmException
- Testing that the JWT header has all the required claims
- Testing that the JWT payload has the required claims
- Testing that user defined payload claims are present in the created JWT

##### Getting JWKs
- Testing that the list of JWKs before JWT creation is empty
- Testing that the list of JWKs after JWT creation is not empty
- Testing that the list of JWKs after JWT creation contains a JWK with the same key id as the `kid` claim in the JWT header
- Testing that all the JWKs in the list can be used to create a valid public key (using modulus and exponent)
- Testing that the public key created from the JWK can be correctly used for JWT verification


### API Tests

##### `/recipe/jwt POST`
- Test that the API throws 400 when the required parameters are missing
- Test that the API returns unsupported algorithm status when an invalid `algorithm` parameter is provided
- Test that the API throws 400 when a negative value for `validity` is provided
- Test that calling the API with a valid request body works fine
- Test that the returned JWT has a valid header
- Test that the returned JWT contains all the user defined payload claims

##### `/recipe/jwt/jwks GET`
- Test that the returned keys list is empty before JWT creation
- Test that the returned keys list is not empty after JWT creation
- Test that the returned list after JWT creation has a key with the same key id as the `kid` claim in the JWT header
- Test that a JWK can be used correctly to verify the returned JWT


# Tests/Changes pending
- [ ] Test to simulate a race condition when writing keys to DB to verify that DuplicateKeyId exception is throws
- [ ] Verify the test for negative validity values. Currently assertion errors are throws as 500 but this error should be handled as a 400

src/main/java/io/supertokens/jwt/JWTSigningKey.java
rishabhpoddar: No i meant not to make it into a function, but to structure the code in a way where you need to run this logic just once per transaction - this is cause we can handle the DuplicateKeyId exception properly, just once. It's not really a big deal as such.. but just easier to read the code in the way I am thinking (I feel)

rishabhpoddar (changes requested):

src/main/java/io/supertokens/inmemorydb/queries/JWTSigningQueries.java
rishabhpoddar: But can't the constructor of the `JWTAsymmetricSigningKeyInfo` always just do that? You can have two constructors in that class - one that accepts just a string, and one that accepts the public and private key (same as the existing constructor)

src/main/java/io/supertokens/jwt/JWTSigningKey.java
rishabhpoddar: I disagree. I think the storage layer should return an empty array. It keeps the types clean (the fewer variations, the better). You should make sure that the storage layer returns an empty list in this case.

src/main/java/io/supertokens/jwt/JWTSigningKey.java
rishabhpoddar: Your last comment is untrue based on our experimentation and tests :) (Just keeping this comment here in case we read the above in the future..)

src/main/java/io/supertokens/inmemorydb/queries/JWTSigningQueries.java
nkshah2: I can do that as well, I just prefer having static methods. But for the sake of consistency probably a better idea to go with multiple constructors, ill make the change

src/main/java/io/supertokens/jwt/JWTSigningKey.java
nkshah2: So thinking about it a little more I think you're right, ill make the change

src/main/java/io/supertokens/inmemorydb/queries/JWTSigningQueries.java
nkshah2: Done (Updated plugin interface PR https://github.com/supertokens/supertokens-plugin-interface/pull/22)

src/main/java/io/supertokens/jwt/JWTSigningKey.java
nkshah2: Done

src/main/java/io/supertokens/jwt/JWTSigningKey.java
nkshah2: Done, im still leaving the logic in the individual method because I think it makes the larger method look a little cleaner but I can change that back if you prefer

src/main/java/io/supertokens/inmemorydb/Start.java
```diff
@@ -804,31 +804,24 @@ public long getThirdPartyUsersCount() throws StorageQueryException {
...
-        }
-    }
+            JWTSigningQueries.setJWTSigningKeyInfo_Transaction(this, sqlCon, info);
+        } catch (StorageTransactionLogicException e) {
 
-    @Override
-    public JWTSigningKeyInfo getLatestJWTSigningKeyForAlgorithm_Transaction(TransactionConnection con, String algorithm)
-            throws StorageQueryException {
-        Connection sqlCon = (Connection) con.getConnection();
+            if (e.actualException.getMessage().equals("[SQLITE_CONSTRAINT]  Abort due to constraint violation (UNIQUE constraint failed: "+ Config.getConfig(this).getJWTSigningKeysTable() +".key_id)")) {
```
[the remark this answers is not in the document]
nkshah2: Done

rishabhpoddar (approved):

src/main/java/io/supertokens/jwt/JWTSigningKey.java
rishabhpoddar: I prefer if it's not in a function.

src/main/java/io/supertokens/jwt/JWTSigningKey.java
```diff
@@ -0,0 +1,238 @@
...
+
+                throw e;
+            }
+        } else if (storage.getType() == STORAGE_TYPE.NOSQL_1) {
+            SessionNoSQLStorage_1 noSQLStorage = (SessionNoSQLStorage_1) storage;
+
+            JWTSigningKeyInfo keyInfo = null;
+
+            while (true) {
+                List<JWTSigningKeyInfo> keysFromStorage = noSQLStorage.getJWTSigningKeys_Transaction();
```
rishabhpoddar, line 164: We should discuss how NoSQL logic works. It seems incorrect on a quick glance

rishabhpoddar (changes requested):

src/main/java/io/supertokens/jwt/JWTSigningKey.java
```diff
@@ -0,0 +1,238 @@
...
+            }
+        }
+    }
+
+    // No SQL storage
+    private JWTSigningKeyInfo generateAndSetKeyInStorage(SessionNoSQLStorage_1 storage, SupportedAlgorithms algorithm) throws StorageTransactionLogicException, UnsupportedJWTSigningAlgorithmException, StorageQueryException {
+        while (true) {
+            try {
+                JWTSigningKeyInfo keyInfo = generateKeyForAlgorithm(algorithm);
+                boolean success = storage.setJWTSigningKeyInfo_Transaction(keyInfo);
```
rishabhpoddar, line 224: The query for the set should be something like insert keyInfo into .. if there is no entry with the matching algo (we need to do this in an atomica way)

rishabhpoddar (changes requested):

src/main/java/io/supertokens/jwt/JWTSigningKey.java
nkshah2: Done

src/main/java/io/supertokens/jwt/JWTSigningKey.java
nkshah2: As discussed we will use findOneAndUpdate in the case of mongo for this

src/main/java/io/supertokens/jwt/JWTSigningKey.java
```diff
@@ -0,0 +1,238 @@
...
+               sqlStorage.commitTransaction(con);
+               return keys;
+            });
+        } else if (storage.getType() == STORAGE_TYPE.NOSQL_1) {
+            SessionNoSQLStorage_1 noSQLStorage = (SessionNoSQLStorage_1) storage;
+
+            List<JWTSigningKeyInfo> keys;
+            List<JWTSigningKeyInfo> keysFromStorage = noSQLStorage.getJWTSigningKeys_Transaction();
+
+            if (keysFromStorage == null) {
```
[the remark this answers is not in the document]
nkshah2: Done

src/main/java/io/supertokens/jwt/JWTSigningKey.java
nkshah2: As discussed we will use findOneAndUpdate in the case of mongo for this

nkshah2:
Status of all tests:
The test that failed was `testThatCoreDriverInterfaceSupportedVersionsAreBeingReturnedByTheAPI` which is expected since the coreDriverInterfaceSupported.json has changed

Screenshot 2021-09-08 at 6 26 31 PM

src/main/java/io/supertokens/jwt/JWTSigningKey.java
```diff
@@ -142,7 +131,17 @@ public JWTSigningKeyInfo getOrCreateAndGetKeyForAlgorithm(SupportedAlgorithms al
 
                     // If no key was found create a new one
                     if (keyInfo == null) {
-                        keyInfo = generateAndSetKeyInStorage(sqlStorage, con, algorithm);
+                        while (true) {
+                            try {
+                                keyInfo = generateKeyForAlgorithm(algorithm);
+                                sqlStorage.setJWTSigningKey_Transaction(con, keyInfo);
+                                return keyInfo;
```
rishabhpoddar, line 138: No return.. We need to break since we are doing commit below.

src/main/java/io/supertokens/jwt/JWTSigningKey.java
```diff
@@ -174,10 +173,21 @@ public JWTSigningKeyInfo getOrCreateAndGetKeyForAlgorithm(SupportedAlgorithms al
...
+                    while (true) {
+                        try {
+                            keyInfo = generateKeyForAlgorithm(algorithm);
+                            boolean success = noSQLStorage.setJWTSigningKeyInfoIfNoKeyForAlgorithmExists_Transaction(keyInfo);
+
+                            if (!success) {
+                                continue;
+                            }
+
+                            return keyInfo;
```
rishabhpoddar, line 185: No return.. we need to break

rishabhpoddar (changes requested):

rishabhpoddar:
> The test that failed was testThatCoreDriverInterfaceSupportedVersionsAreBeingReturnedByTheAPI which is expected since the coreDriverInterfaceSupported.json has changed

So fix the test please.

src/main/java/io/supertokens/jwt/JWTSigningKey.java
rishabhpoddar: Given this bug, i wonder why your tests pass. I think the transaction logic commits on its own if we don't call the commit? Or is it that your tests don't test this case?

src/main/java/io/supertokens/jwt/JWTSigningKey.java
nkshah2: Done

src/main/java/io/supertokens/jwt/JWTSigningKey.java
nkshah2: The transaction logic auto commits but fixed it regardless

nkshah2:
All current and newly added tests pass

Screenshot 2021-09-08 at 9 02 34 PM

rishabhpoddar (approved):

rishabhpoddar:
@nkshah2 we also need to add a test to make sure that the nosql transaction logic is working as expected.

src/test/java/io/supertokens/test/jwt/api/JWTSigningAPITest.java
```diff
@@ -0,0 +1,341 @@
+/*
```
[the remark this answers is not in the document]
nkshah2: Done

src/test/java/io/supertokens/test/jwt/api/JWTSigningAPITest.java
```diff
@@ -0,0 +1,341 @@
...
+        TestingProcessManager.TestingProcess process = TestingProcessManager.start(args);
+        assertNotNull(process.checkOrWaitForEvent(ProcessState.PROCESS_STATE.STARTED));
+
+        JsonObject requestBody = new JsonObject();
+        requestBody.addProperty("jwksDomain", "http://localhost");
+        requestBody.add("payload", new JsonObject());
+        requestBody.addProperty("validity", 3600);
+
+        try {
+            JsonObject response = io.supertokens.httpRequest.HttpRequest
```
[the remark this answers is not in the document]
nkshah2: Done, also renamed the class to HttpRequestForTesting

rishabhpoddar (approved):

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 - New config values `password_reset_token_lifetime`
   and `email_verification_token_lifetime`: https://github.com/supertokens/supertokens-core/issues/297
+- New JWT recipe to create JWT tokens using SuperTokens
```
rishabhpoddar, line 14: Update the changelog to specify db schema changes

rishabhpoddar (changes requested):

CHANGELOG.md
nkshah2: Done

src/test/java/io/supertokens/test/jwt/storage/JWTKeyStorageTest.java
```diff
@@ -0,0 +1,271 @@
...
+            boolean successForSameObj = noSQLStorage_1.setJWTSigningKeyInfoIfNoKeyForAlgorithmExists_Transaction(keyWithDifferentKeyString);
+            assert !successForSameObj;
+
+            try {
+                /*
+                 * When setting with the same key id but different algorithm, findOneAndUpdate will try to insert but Mongo
+                 * should throw an error because the _id of the new document already exists in storage
+                 */
+                JWTSigningKeyInfo keyToSet2 = new JWTSymmetricSigningKeyInfo("keyId-1234", 1000, "EC", "somekeystring");
+                noSQLStorage_1.setJWTSigningKeyInfoIfNoKeyForAlgorithmExists_Transaction(keyToSet2);
```
rishabhpoddar, line 119: Add a fail inside the try / catch

src/test/java/io/supertokens/test/jwt/storage/JWTKeyStorageTest.java
```diff
@@ -0,0 +1,271 @@
...
+
+        for (int i = 0; i < 10; i++) {
+            String algorithm = "alg" + i;
+            String keyId = "key" + i;
+
+            JWTRecipeStorage storage = StorageLayer.getJWTRecipeStorage(process.getProcess());
+
+            if (storage.getType() != STORAGE_TYPE.NOSQL_1) {
+                return;
+            }
```
rishabhpoddar, line 147: Add comment as to why this is being done.

src/test/java/io/supertokens/test/jwt/storage/JWTKeyStorageTest.java
```diff
@@ -0,0 +1,271 @@
...
+                                    boolean success = noSQLStorage_1.setJWTSigningKeyInfoIfNoKeyForAlgorithmExists_Transaction(new JWTSymmetricSigningKeyInfo(keyId + "-1", 1000, algorithm, "SomeKeyString"));
+
+                                    if (!success) {
+                                        fail();
+                                    }
+
+                                    endValueOfCon1.set(keyId + "-1");
+                                } else {
+                                    endValueOfCon1.set(matchingKeyInfo.keyId);
+                                }
```
rishabhpoddar, line 189: Remove this if / else.

src/test/java/io/supertokens/test/jwt/storage/JWTKeyStorageTest.java
```diff
@@ -0,0 +1,271 @@
...
+
+                                break;
+                            } catch (DuplicateKeyIdException e) {
+                                continue;
```
rishabhpoddar, line 193: Throw an exception here.

src/test/java/io/supertokens/test/jwt/storage/JWTKeyStorageTest.java
```diff
@@ -0,0 +1,271 @@
...
+                                // Ignore
+                            }
+
+                            try {
+                                if (matchingKeyInfo == null) {
+                                    boolean success = noSQLStorage_1.setJWTSigningKeyInfoIfNoKeyForAlgorithmExists_Transaction(new JWTSymmetricSigningKeyInfo(keyId + "-2", 1000, algorithm, "SomeKeyString"));
+
+                                    if (!success) {
+                                        continue;
+                                    }
```
rishabhpoddar, line 237: else fail test

src/test/java/io/supertokens/test/jwt/storage/JWTKeyStorageTest.java
```diff
@@ -0,0 +1,271 @@
...
+
+                                    endValueOfCon2.set(keyId + "-2");
+                                } else {
+                                    endValueOfCon2.set(matchingKeyInfo.keyId);
+                                }
+
+                                break;
+                            } catch (DuplicateKeyIdException e) {
+                                continue;
```
rishabhpoddar, line 246: Throw an error instead of continue

src/test/java/io/supertokens/test/jwt/storage/JWTKeyStorageTest.java
```diff
@@ -0,0 +1,271 @@
...
+                            }
+                        }
+                    } catch (Exception ignored) {
+
+                    }
```
rishabhpoddar, line 251: Why do we need this try / catch?

src/test/java/io/supertokens/test/jwt/storage/JWTKeyStorageTest.java
```diff
@@ -0,0 +1,271 @@
...
```
rishabhpoddar, line 198: Why do we need this try / catch?

src/test/java/io/supertokens/test/jwt/storage/JWTKeyStorageTest.java
```diff
@@ -0,0 +1,271 @@
...
+
+    /**
+     * For NoSQL only
+     *
+     * Simulate a race condition for parallel threads trying to read from the database (which results in no rows being found initially)
+     * and writing to storage and make sure that at the end of execution both threads used the same key and only one thread was
+     * able to write to storage
+     */
+    @Test
+    public void testThatSettingJWTKeysInParallelWorksAsExpectedForNoSQL() throws Exception {
```
rishabhpoddar, line 134: You need to generate two different keys with the same algorithm within each thread right before calling the mongodb set function.

rishabhpoddar (changes requested):

src/test/java/io/supertokens/test/jwt/storage/JWTKeyStorageTest.java
nkshah2: Done

src/test/java/io/supertokens/test/jwt/storage/JWTKeyStorageTest.java
nkshah2: Done, added to the test description above the method declaration

src/test/java/io/supertokens/test/jwt/storage/JWTKeyStorageTest.java
nkshah2: Done

src/test/java/io/supertokens/test/jwt/storage/JWTKeyStorageTest.java
nkshah2: Done, the test has been refactored. All exceptions that are not expected result in the test failing.

src/test/java/io/supertokens/test/jwt/storage/JWTKeyStorageTest.java
nkshah2: Done

src/test/java/io/supertokens/test/jwt/storage/JWTKeyStorageTest.java
nkshah2: Done

src/test/java/io/supertokens/test/jwt/storage/JWTKeyStorageTest.java
nkshah2: This is because the calls to storage can throw exceptions (StorageQueryException for example), the test has been refactored to result in failure if any unexpected exceptions occur

src/test/java/io/supertokens/test/jwt/storage/JWTKeyStorageTest.java
nkshah2: Same as above

src/test/java/io/supertokens/test/jwt/storage/JWTKeyStorageTest.java
nkshah2: Done

rishabhpoddar (approved):

rishabhpoddar:
Generate one key of each algo type on core startup.

nkshah2:
Summary of changes in the last couple commits:
- Added logic to Main.java to generate one key for each supported algorithm on startup
- Make algorithm type and key type in the JWK json forced to all caps (this is because most popular JWT verification libraries seem to expect it this way)
- Modify JWKSTest to account for new startup key generation logic
- Removed `JWKSAPITest2_9.testGettingKeysBeforeCreatingJWTReturnsEmpty` in favour of `JWKSTest.testThatThereAreTheSameNumberOfJWKSAsSupportedAlgorithmsBeforeJWTCreation`

src/main/java/io/supertokens/Main.java
```diff
@@ -176,6 +179,18 @@ private void init() throws IOException {
         AccessTokenSigningKey.init(this);
         RefreshTokenKey.init(this);
 
+        // init JWT signing keys, we create one key for each supported algorithm type
+        for (int i = 0; i < JWTSigningKey.SupportedAlgorithms.values().length; i++) {
+            JWTSigningKey.SupportedAlgorithms currentAlgorithm = JWTSigningKey.SupportedAlgorithms.values()[i];
+            try {
+                JWTSigningKey.getInstance(this).getOrCreateAndGetKeyForAlgorithm(currentAlgorithm);
+            } catch (StorageQueryException | StorageTransactionLogicException e) {
+                // We ignore this because in this case the only time
```
rishabhpoddar, line 188: Unfinished comment?

rishabhpoddar (changes requested):

src/main/java/io/supertokens/Main.java
nkshah2: Done

rishabhpoddar (approved):

Issue #305, opened 2021-09-02
Make changing of jwt signing key more smooth


rishabhpoddar:
When the older signing key has changed, use the previous one for some time (2x access token's lifetime?) - so that when the older access tokens continue to function fine and there is no spike in the refresh API.

That being said, any new access token would be created using the new key.

---------
Also, change the default time for jwt signing key change to be 1 week?

rishabhpoddar:
Released.

PR #307, opened 2021-09-09
Multiple access token signing keys

porcellus:
## Summary of change
Made the transition between access token signing keys smoother, by accepting tokens signed by the old one.
We only accept tokens signed by the old key for twice the token validity after the new one has been generated

## Related issues
- supertokens/supertokens-core#305

## Test Plan
Updated tests, plus we may need new cases as well (e.g.: test to see if request spike is gone after signing key change)

## Documentation changes
- [x] Update create tables
- [x] Check documentation about updating signing keys
- [x] Update migration docs
- [ ] Update SaaS config to make default access token signing key interval 7 days
- [ ] Update docs to make default access token signing key interval 7 days 

   - In `build.gradle`

- [x] Documentation
- [ ] Tests


src/main/java/io/supertokens/session/accessToken/AccessTokenSigningKey.java
```diff
@@ -69,142 +73,206 @@ public static AccessTokenSigningKey getInstance(Main main) {
...
             return System.currentTimeMillis() + (10L * 365 * 24 * 3600 * 1000);
         }
     }
 
-    private KeyInfo maybeGenerateNewKeyAndUpdateInDb() throws StorageQueryException, StorageTransactionLogicException {
+    private List<KeyInfo> maybeGenerateNewKeyAndUpdateInDb() throws StorageQueryException, StorageTransactionLogicException {
         Storage storage = StorageLayer.getSessionStorage(main);
         CoreConfig config = Config.getConfig(main);
+        
+        final long validityDuration = config.getAccessTokenSigningKeyUpdateInterval() + 2 * config.getAccessTokenValidity();
```
rishabhpoddar, line 129: - Store the "2" in a constant somewhere on top of this class (as a `final static` variable).
- Also, should we make this configurable by the user? Cause if someone has an app where most people access it at a certain time of the day, and that time is such that it is greater than `validityDuration + timeLatesKeyWasCreated`, there would still be a spike in the refresh APIs.

src/main/java/io/supertokens/session/accessToken/AccessTokenSigningKey.java
```diff
@@ -69,142 +73,206 @@ public static AccessTokenSigningKey getInstance(Main main) {
...
+        final long keyCurrentAfter = System.currentTimeMillis() - config.getAccessTokenSigningKeyUpdateInterval();
+        final long keysValidAfter = keyCurrentAfter - 2 * config.getAccessTokenValidity();
```
rishabhpoddar, line 131: Need better variable names, of comments to explain the purpose of these variables.

src/main/java/io/supertokens/session/accessToken/AccessTokenSigningKey.java
```diff
@@ -69,142 +73,206 @@ public static AccessTokenSigningKey getInstance(Main main) {
...
+            long createdAtTime = this.validKeys.get(0).createdAtTime;
             return createdAtTime + Config.getConfig(main).getAccessTokenSigningKeyUpdateInterval();
         } else {
             // return 10 years from now
             return System.currentTimeMillis() + (10L * 365 * 24 * 3600 * 1000);
         }
     }
 
-    private KeyInfo maybeGenerateNewKeyAndUpdateInDb() throws StorageQueryException, StorageTransactionLogicException {
+    private List<KeyInfo> maybeGenerateNewKeyAndUpdateInDb() throws StorageQueryException, StorageTransactionLogicException {
```
rishabhpoddar, line 125: I think it may make the logic simpler if we decouple the deleting of keys from this method. We can do this by creating a cronjob to delete older keys, and in this method, we fetch all keys and filter the list to remove the keys that are older than `keysValidAfter`.

src/main/java/io/supertokens/session/accessToken/AccessTokenSigningKey.java
```diff
@@ -69,142 +73,206 @@ public static AccessTokenSigningKey getInstance(Main main) {
...
-                if (key != null) {
-                    if (config.getAccessTokenSigningKeyDynamic() && System.currentTimeMillis() > key.createdAtTime
-                            + config.getAccessTokenSigningKeyUpdateInterval()) {
-                        generateNewKey = true;
+            validKeys = sqlStorage.startTransaction(con -> {
+                List<KeyInfo> validKeysFromSQL = new ArrayList<KeyInfo>();
+
+                KeyValueInfo[] keysFromStorage = sqlStorage.getAccessTokenSigningKeys_Transaction(con);
+                
+                boolean generateNewKey = true;
```
rishabhpoddar, line 145: Maybe you don't need the `generateNewKey` variable because instead of doing `if (generateNewKey) {..}`, you can do `if (validKeysFromSQL.size() == 0) {..}`

src/main/java/io/supertokens/session/accessToken/AccessTokenSigningKey.java
```diff
@@ -69,142 +73,206 @@ public static AccessTokenSigningKey getInstance(Main main) {
...
+                boolean clearKeys = false;
+
+                if (keysFromStorage.length == 0) {
```
rishabhpoddar, line 148: I feel that this block can be refactored to be much cleaner.

src/main/java/io/supertokens/session/accessToken/AccessTokenSigningKey.java
```diff
@@ -69,142 +73,206 @@ public static AccessTokenSigningKey getInstance(Main main) {
...
+                    if (legacyKey != null && config.getAccessTokenSigningKeyDynamic()) {
+                        if (keyCurrentAfter <= legacyKey.createdAtTime) {
+                            generateNewKey = false;
+                            validKeysFromSQL.add(new KeyInfo(legacyKey.value, legacyKey.createdAtTime, validityDuration));
+                        } else if (legacyKey.createdAtTime < keysValidAfter) {
+                            sqlStorage.removeLegacyAccessTokenSigningKey_Transaction(con);
+                        }
+                    }
+                } else {
+                    for (KeyValueInfo key: keysFromStorage) {
```
rishabhpoddar, line 160: You may want to consider the legacy key here too?

src/main/java/io/supertokens/session/accessToken/AccessTokenSigningKey.java
```diff
@@ -69,142 +73,206 @@ public static AccessTokenSigningKey getInstance(Main main) {
...
-                }
-
-                boolean generateNewKey = false;
-
-                if (key != null) {
-                    if (config.getAccessTokenSigningKeyDynamic() && System.currentTimeMillis() > key.createdAtTime
-                            + config.getAccessTokenSigningKeyUpdateInterval()) {
-                        generateNewKey = true;
+                validKeys = new ArrayList<KeyInfo>();
+                long lastCreated = -1;
```
rishabhpoddar, line 199: I don't quite understand why we need `lastCreated`

src/main/java/io/supertokens/session/accessToken/AccessTokenSigningKey.java
```diff
@@ -69,142 +73,206 @@ public static AccessTokenSigningKey getInstance(Main main) {
...
+        return Collections.unmodifiableList(validKeys);
     }
 
     public static class KeyInfo {
         public String value;
         public long createdAtTime;
+        public long expiryTime;
 
-        KeyInfo(String value, long createdAtTime) {
+        KeyInfo(String value, long createdAtTime, long validityDuration) {
```
rishabhpoddar, line 272: Why do we need to store `validityDuration` in this object?

src/main/java/io/supertokens/session/accessToken/AccessTokenSigningKey.java
rishabhpoddar, line 125: Let's discuss the `maybeGenerateNewKeyAndUpdateInDb` function over a call.

rishabhpoddar (changes requested):
A few tests to add:
- [ ] Transitioning from the older signing key method to the newer one. For this, you will probably have to make code changes to allow "deactivating" / "activating" of the newer method
- [ ] The ordering of the validKeys is always that the 0th key is the latest one (i.e. issued the most recently)

-----------------
Important edge cases to think / test / write about:
- What would happen if the access token's lifetime is increased / decreased (over a period of time)?
- What would happen if the signing key update interval is increased / decreased (over a period of time)?

src/main/java/io/supertokens/config/CoreConfig.java
```diff
@@ -57,7 +57,7 @@
     private boolean access_token_signing_key_dynamic = true;
 
     @JsonProperty
-    private double access_token_signing_key_update_interval = 24; // in hours
+    private double access_token_signing_key_update_interval = 168; // in hours
```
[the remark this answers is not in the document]
porcellus: Neither is necessary now that we accept tokens signed with older signing keys. Before this PR, the effective lifetime of the access token always ended with the next issued signing, but now it's always the configured lifetime.
Still, we could have the first one.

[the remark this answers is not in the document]
rishabhpoddar: Yea, fair enough. We don't need either of the checks. Resolving this issue.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 - New config values `password_reset_token_lifetime`
   and `email_verification_token_lifetime`: https://github.com/supertokens/supertokens-core/issues/297
+- Added support for multiple access token signing keys: https://github.com/supertokens/supertokens-core/issues/305
```
rishabhpoddar, line 14: Update the changelog to specify db schema changes

rishabhpoddar (changes requested):

src/main/java/io/supertokens/session/accessToken/AccessTokenSigningKey.java
porcellus: It isn't strictly necessary, but we need this info when constructing API responses it's already calculated here.

CHANGELOG.md
rishabhpoddar: In your message for this, please state the name of the table that was added.

src/main/java/io/supertokens/Main.java
```diff
@@ -190,6 +191,11 @@ private void init() throws IOException {
             Cronjobs.addCronjob(this, Telemetry.getInstance(this));
         }
 
+        // starts CleanExpiredAccessTokenSigningKeys cronjob if the access token signing keys can change
+        if (!Config.getConfig(this).getAccessTokenSigningKeyDynamic()) {
```
rishabhpoddar, line 195: Shouldn't this be `if (Config.getConfig(this).getAccessTokenSigningKeyDynamic()) {`? (remove the `!`)

src/main/java/io/supertokens/cronjobs/cleanExpiredAccessTokenSigningKeys/CleanExpiredAccessTokenSigningKeys.java
```diff
@@ -0,0 +1,71 @@
+/*
```
rishabhpoddar, line 1: Consider renaming this to `deleteExpired...` instead of `cleanExpired...` - to be consistent with the names of the other cronjob.

src/main/java/io/supertokens/session/accessToken/AccessTokenSigningKey.java
```diff
@@ -90,29 +96,80 @@ synchronized void removeKeyFromMemoryIfItHasNotChanged(List<KeyInfo> oldKeyInfo)
...
+            SessionSQLStorage sqlStorage = (SessionSQLStorage) storage;
+
+            // start transaction
+            sqlStorage.startTransaction(con -> {
+                KeyValueInfo legacyKey = sqlStorage.getLegacyAccessTokenSigningKey_Transaction(con);
+
+                if (legacyKey != null) {
+                    sqlStorage.addAccessTokenSigningKey_Transaction(con, legacyKey);
+                    sqlStorage.removeLegacyAccessTokenSigningKey_Transaction(con);
+                }
```
rishabhpoddar, line 113: Need to commit transaction. This happens automatically, but still, an explicit one is better anyway.

src/main/java/io/supertokens/session/accessToken/AccessTokenSigningKey.java
```diff
@@ -90,29 +96,80 @@ synchronized void removeKeyFromMemoryIfItHasNotChanged(List<KeyInfo> oldKeyInfo)
...
+            });
+        } else {
+            while (true) {
+                SessionNoSQLStorage_1 noSQLStorage = (SessionNoSQLStorage_1) storage;
+                KeyValueInfoWithLastUpdated legacyKey = noSQLStorage.getLegacyAccessTokenSigningKey_Transaction();
+
+                if (legacyKey != null) {
+                    // We should only get here once, after an upgrade.
+                    // We avoid adding duplicates by enforcing that a legacy key can only ever be
+                    // the first one in the new table.
```
rishabhpoddar, line 124: How are we enforcing this? By removing the legacy key post update right? Or some other trick?

src/main/java/io/supertokens/session/accessToken/AccessTokenSigningKey.java
```diff
@@ -90,29 +96,80 @@ synchronized void removeKeyFromMemoryIfItHasNotChanged(List<KeyInfo> oldKeyInfo)
...
+                KeyValueInfo legacyKey = sqlStorage.getLegacyAccessTokenSigningKey_Transaction(con);
+
+                if (legacyKey != null) {
+                    sqlStorage.addAccessTokenSigningKey_Transaction(con, legacyKey);
+                    sqlStorage.removeLegacyAccessTokenSigningKey_Transaction(con);
+                }
+                return legacyKey;
+            });
+        } else {
+            while (true) {
```
rishabhpoddar, line 117: Why do we need to put this in a while true loop?

src/main/java/io/supertokens/session/accessToken/AccessTokenSigningKey.java
rishabhpoddar: Aren't we checking the transaction thing by reading the first item of the array in mongo and seeing it's createdTime? - this is why I asked if we need `lastCreated` or not.

src/test/java/io/supertokens/test/session/CleanExpiredAccessTokenSigningKeysTest.java
```diff
@@ -0,0 +1,63 @@
...
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TestRule;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+
+public class CleanExpiredAccessTokenSigningKeysTest {
```
rishabhpoddar, line 34: You need to add a test in which there are a few keys in the db, some of which are expired, the cronjob runs, and then removes just the expired keys.

rishabhpoddar (changes requested):

rishabhpoddar:
We should discuss the following:
- [ ] Someone who is running the older node SDK (2.8 CDI version), using the new core.
- [ ] Someone using the new node SDK with the older core (2.8 CDI version).
- [ ] Someone having multiple signing keys in the db, and then they switch off dynamic update of signing keys.
- [ ] Implications of switching of dynamic update of signing keys (from the core's and node's perspective)

src/main/java/io/supertokens/session/accessToken/AccessTokenSigningKey.java
porcellus: We are setting lastCreated to 0, which means that there is no other key.

src/main/java/io/supertokens/session/accessToken/AccessTokenSigningKey.java
porcellus: No need, it was part of a transaction with retry originally. I'll remove it.

src/main/java/io/supertokens/session/accessToken/AccessTokenSigningKey.java
porcellus: It's actually better to set it to -1, as it better signals the intent.

CHANGELOG.md
porcellus: Should I add the table name here? It feels a bit out of place since it's different in Mongo.

src/main/java/io/supertokens/session/accessToken/AccessToken.java
```diff
@@ -46,66 +48,79 @@ private static AccessTokenInfo getInfoFromAccessToken(@Nonnull Main main, @Nonnu
...
-        Utils.PubPriKey signingKey = new Utils.PubPriKey(keyInfo.value);
-        try {
-            final JWT.JWTInfo jwtInfo;
+        
+        List<AccessTokenSigningKey.KeyInfo> keyInfoList = AccessTokenSigningKey.getInstance(main).getAllKeys();
+
+        Exception error = null;
+        JWT.JWTInfo jwtInfo = null;
+        for (KeyInfo keyInfo: keyInfoList) {
+            Utils.PubPriKey signingKey = new Utils.PubPriKey(keyInfo.value);
```
rishabhpoddar, line 57: You should check the signing key's expiry time here too.

rishabhpoddar (changes requested):

CHANGELOG.md
rishabhpoddar: Just say something like "Added a new table named ... for SQL storage"

src/main/java/io/supertokens/session/accessToken/AccessTokenSigningKey.java
rishabhpoddar: You can make the plugin interface function take a `Long`. This way, you can give the function `null` instead of `-1` or `0`

src/main/java/io/supertokens/session/accessToken/AccessTokenSigningKey.java
```diff
@@ -90,29 +96,80 @@ synchronized void removeKeyFromMemoryIfItHasNotChanged(List<KeyInfo> oldKeyInfo)
...
+                SessionNoSQLStorage_1 noSQLStorage = (SessionNoSQLStorage_1) storage;
+                KeyValueInfoWithLastUpdated legacyKey = noSQLStorage.getLegacyAccessTokenSigningKey_Transaction();
+
+                if (legacyKey != null) {
+                    // We should only get here once, after an upgrade.
+                    // We avoid adding duplicates by enforcing that a legacy key can only ever be
+                    // the first one in the new table.
+                    if (noSQLStorage.addAccessTokenSigningKey_Transaction(
+                            new KeyValueInfo(legacyKey.value, legacyKey.createdAtTime), 0)) {
```
rishabhpoddar: Ignore the result of the insertion..

rishabhpoddar (changes requested):

porcellus:
- Someone having multiple signing keys in the db, and then they switch off dynamic update of signing keys.
=> In this case, the signing keys will be left in the database, and tokens signed by them are accepted. The core should use the latest key to sign new tokens, so the older ones will not be used after their tokens have expired.

- Implications of switching OFF dynamic update of signing keys (from the core's and node's perspective)
=> Switching off the dynamic update causes the current set of keys (expired or not) to stay in the database. This should make no difference to the node SDK: token verification should succeed until the expiration time.

- Implications of switching ON dynamic update of signing keys (from the core's and node's perspective)
 => Switching on dynamic update makes all issued access tokens expire, causing a one-time spike, like the ones we had before this PR.

rishabhpoddar (changes requested):
> In this case, the signing keys will be left in the database, and tokens signed by them are accepted. The core should use the latest key to sign new tokens, so the older ones will not be used after their tokens have expired.

The older tokens would never expire correct? Cause the `getAccessTokenSigningKeyUpdateInterval()` function would return a really large number.

src/main/java/io/supertokens/config/CoreConfig.java
```diff
@@ -137,7 +137,8 @@ public boolean getAccessTokenSigningKeyDynamic() {
     }
 
     public long getAccessTokenSigningKeyUpdateInterval() {
-        return (long) (access_token_signing_key_update_interval * 3600 * 1000);
+        return access_token_signing_key_dynamic ?
+            (long) (access_token_signing_key_update_interval * 3600 * 1000) : (100L * 365 * 24 * 3600 * 1000);
```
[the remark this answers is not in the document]
porcellus: 1. 10L vs. 100L: updated.
2. Calling `getAccessTokenSigningKeyDynamic`: I'd still prefer to call this explicitly in `cleanExpiredAccessTokenSigningKeys`. It's one less DB call (although it's rare and shouldn't matter), but more importantly, this seems like an important thing to check before removing anything.

src/main/java/io/supertokens/session/accessToken/AccessTokenSigningKey.java
```diff
@@ -69,142 +79,227 @@ public static AccessTokenSigningKey getInstance(Main main) {
...
-                + config.getAccessTokenSigningKeyUpdateInterval())) {
+
+        // If a key in the cache has expired we should refresh the list.
+        if (this.validKeys.stream().anyMatch(((KeyInfo k) -> k.expiryTime < System.currentTimeMillis()))) {
+            this.validKeys = maybeGenerateNewKeyAndUpdateInDb();
+        }
+
+        if (config.getAccessTokenSigningKeyDynamic()
+                && System.currentTimeMillis() > (this.validKeys.get(0).createdAtTime
+                        + config.getAccessTokenSigningKeyUpdateInterval())) {
```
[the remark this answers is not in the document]
porcellus: `maybeGenerateNewKeyAndUpdateInDb` should be called exactly once here, so all 3 if statements could be combined into a single one.
We can filter keys first, but `this.validKeys.size() == 0` should never be true otherwise. If it's simpler that way I'll update it.

porcellus:
I couldn't reply on the other comment, but:

> The older tokens would never expire correct? Cause the getAccessTokenSigningKeyUpdateInterval() function would return a really large number.

They would expire, the token expiration isn't affected by the signing key update interval.

src/test/java/io/supertokens/test/session/AccessTokenTest.java
```diff
@@ -245,7 +245,7 @@ public void signingKeyShortInterval() throws InterruptedException,
     }
 
     @Test
-    public void signingKeyChangeThrowsRefreshTokenError()
+    public void signingKeyChangeDoesNotThrow()
```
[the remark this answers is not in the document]
rishabhpoddar: @porcellus perhaps you missed this comment?

rishabhpoddar:
> They would expire, the token expiration isn't affected by the signing key update interval.

How would that be the case? If access token signing key dynamic is `false`, then the `config.getAccessTokenSigningKeyUpdateInterval()` would return a very large value. Therefore, [this variable](https://github.com/supertokens/supertokens-core/blob/549e37cca6b8fbddeba685ed41a1c91da1ac531a/src/main/java/io/supertokens/session/accessToken/AccessTokenSigningKey.java#L182) would be very large, and I believe that this variable is a basis of deciding if a key has expired? Perhaps I am missing something.

src/test/java/io/supertokens/test/session/DeleteExpiredAccessTokenSigningKeysTest.java
```diff
@@ -0,0 +1,113 @@
...
+            DeleteExpiredAccessTokenSigningKeys.getInstance(process.getProcess()).getIntervalTimeSeconds(),
+            Config.getConfig(process.getProcess()).getAccessTokenSigningKeyUpdateInterval() / 1000
+        );
+
+        process.kill();
+        assertNotNull(process.checkOrWaitForEvent(ProcessState.PROCESS_STATE.STOPPED));
+    }
+    
+    @Test
+    public void jobCleansOldKeysTest() throws Exception {
```
rishabhpoddar, line 71: Have this same time, but with access token signing key dynamic setting set to `false`. And then check that the keys that are removed in this test, don't get removed in that test.

rishabhpoddar (changes requested):

rishabhpoddar:
Also,  you should also add a test to make sure that if a key exists in the key-value table (from older version of the core), and then when the new one is started, it moves the key successfully, and removes it from the older table (if this is not already added)

porcellus:
The `signingKeyLifetime` is the basis for how long we use a key to verify signatures. This means that they pass signature verification, but their expiration is checked in [here](https://github.com/supertokens/supertokens-core/blob/549e37cca6b8fbddeba685ed41a1c91da1ac531a/src/main/java/io/supertokens/session/accessToken/AccessToken.java#L119)

porcellus:
>Also, you should also add a test to make sure that if a key exists in the key-value table (from older version of the core), and then when the new one is started, it moves the key successfully, and removes it from the older table (if this is not already added)

There is a test that does this by calling the migration method instead of starting a new core. Should I add another that (re)starts the core?

src/test/java/io/supertokens/test/session/AccessTokenTest.java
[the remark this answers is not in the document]
porcellus: We talked about this briefly on a call: This is a test case that I felt didn't make sense anymore: signing key change should not cause validation to throw an error, only token expiration should which is a separate test.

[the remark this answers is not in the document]
rishabhpoddar: Alright!

rishabhpoddar (approved):

PR #329, opened 2021-11-12
Fix JWT expiry calculation logic

nkshah2:
## Summary of change
When creating a JWT, the logic for calculating expiry would use long/integer division which would end up rounding down. To fix this we use `double` instead of `long` when calculating the JWT expiry

## Related issues
NA

## Test Plan
All existing tests pass

Screenshot 2021-11-12 at 1 20 51 PM

## Documentation changes
NA

   - In `build.gradle`

- [x] Update version for supertokens-core (patch number)
- [x] Update CHANGELOG


src/main/java/io/supertokens/jwt/JWTSigningFunctions.java
```diff
@@ -83,8 +83,8 @@ public static String createJWTToken(Main main, String algorithm, JsonObject payl
         headerClaims.put("typ", "JWT");
         headerClaims.put("kid", keyToUse.keyId);
 
-        long currentTimeInMillis = System.currentTimeMillis();
-        long jwtExpiry = (currentTimeInMillis / 1000) + (jwtValidity); // JWT Expiry is seconds from epoch not millis
+        double currentTimeInMillis = System.currentTimeMillis();
+        double jwtExpiry = (currentTimeInMillis / 1000) + (jwtValidity); // JWT Expiry is seconds from epoch not millis
```
rishabhpoddar, line 87:
```
long currentTimeInMillis = System.currentTimeMillis();
long jwtExpiry = roundUp((currentTimeInMillis / 1000.0) + (jwtValidity)); // JWT Expiry is seconds from epoch not millis
```

rishabhpoddar: Also need to add test for this - if you can

rishabhpoddar (changes requested):

src/main/java/io/supertokens/jwt/JWTSigningFunctions.java
nkshah2: Changed the logic and all existing tests pass

Issue #333, opened 2021-11-15
Implement passwordless in Core


porcellus:
API, logic, and DB are defined in other issues in this epic.

porcellus:
The initial implementation is in a PR, it will be continued in separate issues (for testing and fixes if necessary).

Issue #322, opened 2021-10-22
Core API logic pseudo code for passwordless


porcellus:
Core API logic pseudo code for passwordless

rishabhpoddar:
@porcellus the CDI and FDI spec need to be updated.

rishabhpoddar:
# Feedback

## https://app.code2flow.com/OFxcbh1FNLXd (POST /recipe/sigininup/code/consume)
- [ ] After we successfully get a lock on line 36, shouldn't we then check if the device exists and the linkCode exists? I guess my point here is that we should try and get a lock on the device_id_hash on the passwordless_devices table as soon as possible. That would be before we check if the code has expired etc.. So the flow would be:
   - If linkCode is passed, we fetch the device_id_hash based on that
   - If device_id is passed, we hash it to get he device_id
   - Now that we have the device_id, we lock it (so in the diagram, there should only be one box for locking)
   - Post the above, check if the device still exists and for code matching / expiry etc...
   - If there is a deadlock, we go back to the step of locking with this device_id.
- [ ] There is a comment that is incomplete "No need to lock, because"
- [ ] Should we let line 75 happen via the cronjob instead of doing it here? Instead here, when we consume a code, we check if it's already exceeded the failed attempts and if it has, we send a restart flow error. I say this cause:
   - If a user changes (by reducing) the config for max number of failed attempts, then there could be existing rows in the db that were invalid before, but now they won't be. So there has to be something that removes them (the cronjob), and this API will have to take this scenario into account.
- [ ] line 61 -> can you explain why in the diagram?

## https://app.code2flow.com/MnqlXSzyjajO (cronjob)
- [ ] "get and lock all expired codes" -> shouldn't we actually be locking on device_id_hash? So this would be get all expired codes' device_ids. Then for each one, we loop through and lock that device_id_hash, and delete its expired code along with the device itself if there are no codes left.
- [ ] Also, this needs to take into account max number of failed attempts being passed (as per my comment above)

porcellus:
> After we successfully get a lock on line 36, shouldn't we then check if the device exists and the `linkCode` exists?

- We should check if the device exists (I kinda forgot I always did that explicitly after each get).
- In user code input flow the device is locked before all of this.
- Checking the code expiration before locking the device doesn't take any time and we save 1 lock if someone keeps clicking on an expired link (if we are doing it in separate queries, we could actually select, lock and check the expiration in a single one)
[...]

PR #338, opened 2021-11-22
Passwordless consume code

porcellus:
## Summary of change
Adding consume code API

## Related issues
- #322 
- #333 

## Test Plan
Added some of the basic testcases the rest will be collected on the passwordless->version branch PR.

### DIFF
--- a/src/main/java/io/supertokens/passwordless/Passwordless.java
+++ b/src/main/java/io/supertokens/passwordless/Passwordless.java
@@ -19,7 +19,6 @@ package io.supertokens.passwordless;
 import java.security.InvalidKeyException;
 import java.security.NoSuchAlgorithmException;
 import java.security.SecureRandom;
-import java.util.Base64;
 import java.util.Objects;
 import java.util.stream.Stream;
 
@@ -135,50 +134,52 @@ public class Passwordless {
         long passwordlessCodeLifetime = Config.getConfig(main).getPasswordlessCodeLifetime();
         int maxCodeInputAttempts = Config.getConfig(main).getPasswordlessMaxCodeInputAttempts();
 
-        String deviceIdHash;
-        String linkCodeHash;
+        PasswordlessDeviceIdHash deviceIdHash;
+        PasswordlessLinkCodeHash linkCodeHash;
         if (linkCode != null) {
-            byte[] linkCodeBytes = Base64.getUrlDecoder().decode(linkCode);
-            linkCodeHash = Utils.hashSHA256Base64(linkCodeBytes);
+            PasswordlessLinkCode parsedCode = PasswordlessLinkCode.decodeString(linkCode);
+            linkCodeHash = parsedCode.getHash();
 
-            PasswordlessCode code = passwordlessStorage.getCodeByLinkCodeHash(linkCodeHash);
+            PasswordlessCode code = passwordlessStorage.getCodeByLinkCodeHash(linkCodeHash.encode());
             if (code == null || code.createdAt < (System.currentTimeMillis() - passwordlessCodeLifetime)) {
                 throw new RestartFlowException();
             }
-            deviceIdHash = code.deviceIdHash;
+            deviceIdHash = new PasswordlessDeviceIdHash(code.deviceIdHash);
         } else {
-            byte[] deviceIdBytes = Base64.getDecoder().decode(deviceId);
-            deviceIdHash = Utils.hashSHA256Base64UrlSafe(deviceIdBytes);
+            PasswordlessDeviceId parsedDeviceId = PasswordlessDeviceId.decodeString(deviceId);
 
-            byte[] linkCodeBytes = Utils.hmacSHA256(deviceIdBytes, userInputCode);
-            linkCodeHash = Utils.hashSHA256Base64(linkCodeBytes);
+            deviceIdHash = parsedDeviceId.getHash();
+            linkCodeHash = parsedDeviceId.getLinkCode(userInputCode).getHash();
         }
 
         PasswordlessDevice consumedDevice;
         try {
             consumedDevice = passwordlessStorage.startTransaction(con -> {
-                PasswordlessDevice device = passwordlessStorage.getDevice_Transaction(con, deviceIdHash);
+                PasswordlessDevice device = passwordlessStorage.getDevice_Transaction(con, deviceIdHash.encode());
                 if (device == null) {
                     throw new StorageTransactionLogicException(new RestartFlowException());
                 }
                 if (device.failedAttempts >= maxCodeInputAttempts) {
-                    passwordlessStorage.deleteDevice_Transaction(con, deviceIdHash);
+                    // This can happen if the configured maxCodeInputAttempts changes
+                    passwordlessStorage.deleteDevice_Transaction(con, deviceIdHash.encode());
                     passwordlessStorage.commitTransaction(con);
                     throw new StorageTransactionLogicException(new RestartFlowException());
                 }
 
-                PasswordlessCode code = passwordlessStorage.getCodeByLinkCodeHash_Transaction(con, linkCodeHash);
+                PasswordlessCode code = passwordlessStorage.getCodeByLinkCodeHash_Transaction(con,
+                        linkCodeHash.encode());
                 if (code == null || code.createdAt < System.currentTimeMillis() - passwordlessCodeLifetime) {
                     if (deviceId != null) {
                         // If we get here, it means that the user tried to use a userInputCode, but it was incorrect or
                         // the code expired. This means that we need to increment failedAttempts or clean up the device
                         // if it would exceed the configured max.
                         if (device.failedAttempts + 1 >= maxCodeInputAttempts) {
-                            passwordlessStorage.deleteDevice_Transaction(con, deviceIdHash);
+                            passwordlessStorage.deleteDevice_Transaction(con, deviceIdHash.encode());
                             passwordlessStorage.commitTransaction(con);
                             throw new StorageTransactionLogicException(new RestartFlowException());
                         } else {
-                            passwordlessStorage.incrementDeviceFailedAttemptCount_Transaction(con, deviceIdHash);
+                            passwordlessStorage.incrementDeviceFailedAttemptCount_Transaction(con,
+                                    deviceIdHash.encode());
                             passwordlessStorage.commitTransaction(con);
 
                             if (code != null) {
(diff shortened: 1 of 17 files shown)


src/main/java/io/supertokens/passwordless/Passwordless.java
```diff
@@ -124,6 +129,121 @@ private static String generateUserInputCode() {
...
+
+        String deviceIdHash;
+        String linkCodeHash;
+        if (linkCode != null) {
+            byte[] linkCodeBytes = Base64.decodeBase64URLSafe(linkCode);
+            byte[] linkCodeHashBytes = Utils.hashSHA256Bytes(linkCodeBytes);
+            linkCodeHash = Base64.encodeBase64String(linkCodeHashBytes);
+
+            PasswordlessCode code = passwordlessStorage.getCodeByLinkCodeHash(linkCodeHash);
+            if (code == null || code.createdAt < System.currentTimeMillis() - passwordlessCodeLifetime) {
```
rishabhpoddar, line 147: Put brackets around `System.currentTimeMillis() - passwordlessCodeLifetime`

src/main/java/io/supertokens/passwordless/Passwordless.java
```diff
@@ -124,6 +129,121 @@ private static String generateUserInputCode() {
...
+        PasswordlessSQLStorage passwordlessStorage = StorageLayer.getPasswordlessStorage(main);
+        long passwordlessCodeLifetime = Config.getConfig(main).getPasswordlessCodeLifetime();
+        int maxCodeInputAttempts = Config.getConfig(main).getPasswordlessMaxCodeInputAttempts();
+
+        String deviceIdHash;
+        String linkCodeHash;
+        if (linkCode != null) {
+            byte[] linkCodeBytes = Base64.decodeBase64URLSafe(linkCode);
+            byte[] linkCodeHashBytes = Utils.hashSHA256Bytes(linkCodeBytes);
+            linkCodeHash = Base64.encodeBase64String(linkCodeHashBytes);
```
rishabhpoddar, line 144: Can extract this into a function since it's being replicated in multiple places.

src/main/java/io/supertokens/passwordless/Passwordless.java
```diff
@@ -124,6 +129,121 @@ private static String generateUserInputCode() {
...
+                throw new RestartFlowException();
+            }
+            deviceIdHash = code.deviceIdHash;
+        } else {
+            byte[] deviceIdBytes = Base64.decodeBase64(deviceId);
+            deviceIdHash = Base64.encodeBase64URLSafeString(Utils.hashSHA256Bytes(deviceIdBytes));
+
+            byte[] linkCodeBytes = Utils.hmacSHA256(deviceIdBytes, userInputCode);
+            byte[] linkCodeHashBytes = Utils.hashSHA256Bytes(linkCodeBytes);
+            linkCodeHash = Base64.encodeBase64String(linkCodeHashBytes);
```
rishabhpoddar, line 157: Can extract this into a function since it's being replicated in multiple places.

src/main/java/io/supertokens/passwordless/Passwordless.java
```diff
@@ -124,6 +129,121 @@ private static String generateUserInputCode() {
...
+                if (device == null) {
+                    throw new StorageTransactionLogicException(new RestartFlowException());
+                }
+                if (device.failedAttempts >= maxCodeInputAttempts) {
+                    passwordlessStorage.deleteDevice_Transaction(con, deviceIdHash);
+                    passwordlessStorage.commitTransaction(con);
+                    throw new StorageTransactionLogicException(new RestartFlowException());
+                }
+
+                PasswordlessCode code = passwordlessStorage.getCodeByLinkCodeHash_Transaction(con, linkCodeHash);
```
rishabhpoddar, line 173: `getCodeByLinkCodeHash_Transaction` should not take a lock inside it. Just confirm that this is true.

src/main/java/io/supertokens/passwordless/Passwordless.java
```diff
@@ -124,6 +129,121 @@ private static String generateUserInputCode() {
...
+                if (code == null || code.createdAt < System.currentTimeMillis() - passwordlessCodeLifetime) {
+                    if (deviceId != null) {
```
rishabhpoddar, line 175: Add a comment explaining that if we come inside this if statement, if means that the user used the user input code and failed. This means that we must increment the failed attempts count as shown below.

src/main/java/io/supertokens/passwordless/Passwordless.java
```diff
@@ -124,6 +129,121 @@ private static String generateUserInputCode() {
...
+                        if (device.failedAttempts >= maxCodeInputAttempts) {
```
rishabhpoddar, line 176: This `if` would never actually be true cause if it were, we would have thrown a restart flow error in the `if` of line 167

src/main/java/io/supertokens/passwordless/Passwordless.java
```diff
@@ -124,6 +129,121 @@ private static String generateUserInputCode() {
...
+            }
+            if (e.actualException instanceof IncorrectUserInputCodeException) {
+                throw (IncorrectUserInputCodeException) e.actualException;
+            }
+            if (e.actualException instanceof RestartFlowException) {
+                throw (RestartFlowException) e.actualException;
+            }
+            throw e;
+        }
+
```
rishabhpoddar, line 217: Add a comment here saying that if the execution comes here, it means that we have successfully consumed the code

src/main/java/io/supertokens/passwordless/Passwordless.java
```diff
@@ -124,6 +129,121 @@ private static String generateUserInputCode() {
...
+                try {
+                    if (user == null) {
+                        String userId = Utils.getUUID();
+                        long timeJoined = System.currentTimeMillis();
+                        user = new UserInfo(userId, consumedDevice.email, consumedDevice.phoneNumber, timeJoined);
+                        passwordlessStorage.createUser(user);
+                        return new ConsumeCodeResponse(true, user);
+                    }
+                } catch (DuplicateEmailException | DuplicatePhoneNumberException e) {
+                    throw new RestartFlowException();
```
rishabhpoddar, line 231: Write why we are throwing `RestartFlowException` instead of signing in the user. I mean an explanation of a scenario in which it would come here and what the UX will be at that time.

src/main/java/io/supertokens/passwordless/Passwordless.java
```diff
@@ -124,6 +129,121 @@ private static String generateUserInputCode() {
...
+                    // We can retry..
+                }
+            }
+        } else {
+            if (user.email != null) {
+                removeCodesByEmail(main, user.email);
+            }
+            if (user.phoneNumber != null) {
+                removeCodesByPhoneNumber(main, user.phoneNumber);
+            }
```
rishabhpoddar, line 242: Why are we doing this only in case the user exists? What if the execution ends up in line number 231 (where we throw a `RestartFlowException`). Then also we want to do this?

rishabhpoddar (changes requested):

src/main/java/io/supertokens/passwordless/Passwordless.java
porcellus: It doesn't take a lock and also includes a comment on why.

src/main/java/io/supertokens/passwordless/Passwordless.java
porcellus: Added comments to the code that hopefully helps explain this. If not, I'm happy to expand on it.

src/test/java/io/supertokens/test/passwordless/api/PasswordlessConsumeCodeAPITest2_10.java
```diff
@@ -286,7 +286,7 @@ public void testExpiredUserInputCode() throws Exception {
     public void testIncorrectUserInputCode() throws Exception {
         String[] args = { "../" };
 
-        Utils.setValueInConfig("passwordless_max_code_input_attempts", "1"); // Only 1 code entry permitted
+        Utils.setValueInConfig("passwordless_max_code_input_attempts", "2"); // Only 1 code entry permitted
```
rishabhpoddar, line 289: The comment here says only 1 entry permitted, but you have set it to 2.

src/main/java/io/supertokens/utils/Utils.java
```diff
@@ -84,6 +84,16 @@ public static String hashSHA256(String base) throws NoSuchAlgorithmException {
...
+    public static String hashSHA256Base64UrlSafe(byte[] base) throws NoSuchAlgorithmException {
+        byte[] hashBytes = Utils.hashSHA256Bytes(base);
+        return Base64.getUrlEncoder().encodeToString(hashBytes);
+    }
+
+    public static String hashSHA256Base64(byte[] base) throws NoSuchAlgorithmException {
+        byte[] hashBytes = Utils.hashSHA256Bytes(base);
+        return Base64.getEncoder().encodeToString(hashBytes);
+    }
+
```
rishabhpoddar, line 96: In my mind, extracting those into functions entail creating helper functions for:
- deviceId -> deviceIdHash
- (deviceId + userInputCode) -> linkCode
- linkCode -> linkCodeHash

And then these functions would be used everywhere as and when needed. This way, we have one central place where do do conversations using base64 and bytes etc..

rishabhpoddar (changes requested):

src/main/java/io/supertokens/passwordless/Passwordless.java
```diff
@@ -124,6 +128,129 @@ private static String generateUserInputCode() {
...
+            consumedDevice = passwordlessStorage.startTransaction(con -> {
+                PasswordlessDevice device = passwordlessStorage.getDevice_Transaction(con, deviceIdHash);
+                if (device == null) {
+                    throw new StorageTransactionLogicException(new RestartFlowException());
+                }
+                if (device.failedAttempts >= maxCodeInputAttempts) {
+                    passwordlessStorage.deleteDevice_Transaction(con, deviceIdHash);
+                    passwordlessStorage.commitTransaction(con);
+                    throw new StorageTransactionLogicException(new RestartFlowException());
+                }
```
rishabhpoddar, line 168: This `if` would never actually be true cause of the `if` in line 176. So I assume that this is here cause we want to take into account the case where the dev might reduce the value of `maxCodeInputAttempts`. Is this true? If so, please add a comment here explaining this - as it's not very obvious.

rishabhpoddar (changes requested):

src/main/java/io/supertokens/utils/Utils.java
porcellus: My preference would be to do the b64 conversions at the API level. It doesn't matter much, but passing the unparsed version around and doing the conversion multiple times doesn't feel clean.

src/main/java/io/supertokens/passwordless/PasswordlessLinkCodeHash.java
```diff
@@ -0,0 +1,22 @@
...
+    public PasswordlessLinkCodeHash(byte[] bytes) {
+        // We never do anything further with the bytes, so we can just encode, store and reuse it.
+        // If we choose to do storage based on bytes this can change.
+        this.encodedValue = Base64.getUrlEncoder().encodeToString(bytes);
+    }
+
+    public PasswordlessLinkCodeHash(String encodedValue) {
+        this.encodedValue = encodedValue;
+        // We could do validation here...
+    }
```
rishabhpoddar, line 17: If this is not being used anywhere, we should remove this. Also, the comment indicates that something is missing.

rishabhpoddar (changes requested):

src/main/java/io/supertokens/passwordless/PasswordlessLinkCodeHash.java
porcellus: I just added that to mirror the one from DeviceIdHash. Validation is indeed something we could do, but I'm not sure if it's necessary.

rishabhpoddar (approved):

PR #340, opened 2021-11-22
Adding API to remove codes by email or phone

porcellus:
## Summary of change
Adding consume remove API

## Related issues
- #322 
- #333 

## Test Plan
Added some of the basic testcases the rest will be collected on the passwordless->version branch PR.


src/main/java/io/supertokens/inmemorydb/queries/PasswordlessQueries.java
```diff
@@ -139,23 +139,42 @@ public static void deleteDevice_Transaction(Start start, Connection con, String
...
+            ResultSet result = pst.executeQuery();
+            while (result.next()) {
+                deviceIdHashes.add(result.getString("device_id_hash"));
+            }
         }
+        deleteRowsByDeviceIdHashList_Transaction(con, Config.getConfig(start).getPasswordlessCodesTable(),
+                deviceIdHashes);
+        deleteRowsByDeviceIdHashList_Transaction(con, Config.getConfig(start).getPasswordlessDevicesTable(),
+                deviceIdHashes);
     }
```
rishabhpoddar, line 158: Why are we doing it this way? Why can't we just do a `DELETE from devices where phoneNumber = ..` (and this will also delete from the codes table automatically)? Cause If we do it this way, then we need to make the SELECT query a `FOR_UPDATE` one?

rishabhpoddar (changes requested):

src/main/java/io/supertokens/inmemorydb/queries/PasswordlessQueries.java
porcellus: I went with locking the entire table - it simplifies implementation, and it shouldn't matter much in trial&dev.

> creating a new device + code

I don't think we need to lock the table when creating a device with code. Creating the code and device is done in a transaction, and there is no pre-condition for creating the device depending on anything else in the DB.

> creating a new code based on a device id hash
> updating info in the devices table

Locks using `getDevice_Transaction`

rishabhpoddar: > I don't think we need to lock the table when creating a device with code. Creating the code and device is done in a transaction, and there is no pre-condition for creating the device depending on anything else in the DB.

yes. For creating a new device and a code, no locking is required. But for creating a new code, given an existing device, we would need to lock the devices table?

src/main/java/io/supertokens/inmemorydb/queries/PasswordlessQueries.java
```diff
@@ -149,23 +151,48 @@ public static void deleteDevice_Transaction(Start start, Connection con, String
...
 
     public static void createCode_Transaction(Start start, Connection con, PasswordlessCode code) throws SQLException {
```
rishabhpoddar, line 198: If this function is not being used anywhere outside this file, either:
- make it a part of the function which is calling it (in case it's just one function)
- Make this `private static void`

src/main/java/io/supertokens/inmemorydb/ConnectionWithLocks.java
```diff
@@ -17,26 +17,28 @@
...
     private ConnectionPool connectionPool;
-    private List<String> lockedKeys = new ArrayList<>();
+    private Set<String> lockedKeys = new HashSet<String>();
 
     public void lock(String key) {
-        connectionPool.lock(key);
-        this.lockedKeys.add(key);
+        if (!this.lockedKeys.contains(key)) {
+            this.lockedKeys.add(key);
+            connectionPool.lock(key);
```
rishabhpoddar, line 35: This is guaranteed to work only of the same thread is going to be calling `lock` on the same instance of `ConnectionWithLocks ` (which I think is true).

If one `ConnectionWithLocks` instance is being shared by multiple threads (which I think never happens), there can still be a deadlock here cause two threads can try and lock the same `key` and get `this.lockedKeys.contains(key)` as `false`, thereby entering the `if` statement, and then both running `connectionPool.lock(key)`, which will hang that thread. In this case, we can solve this by making the `lock` function into `public synchronized void lock(String key) {` and same for `unlockAllLocks` function (it might be worth adding the `synchronized` keyword anyway.

rishabhpoddar (changes requested):

src/main/java/io/supertokens/inmemorydb/ConnectionWithLocks.java
porcellus: Hmm, I didn't look too deep at the rest of the class to see if it'd work while shared between threads. I think this wouldn't be misleading (saying this should work when shared) and it doesn't have too much overhead as far as I can see, so I'll add it.

rishabhpoddar: Alright!

rishabhpoddar (approved):

Issue #381, opened 2022-02-27
User metadata


porcellus:
## 🚀 User metadata

Storing metadata about each user to enable progressive profiling.

### Frontend:
N/A

### Backend:
New recipe, with functions to get, create/update and delete user metadata using the user id.
No new endpoints or changes to the FDI.

The details will be added in a separate issue.

### CDI:
https://github.com/supertokens/core-driver-interface/pull/35

#### PSQL:
```
CREATE TABLE IF NOT EXISTS user_metadata (
    user_id VARCHAR(128) NOT NULL,
    user_metadata TEXT,
    CONSTRAINT user_metadata_pkey PRIMARY KEY (user_id)
);
```
#### MySQL:
```
CREATE TABLE IF NOT EXISTS user_metadata (
    user_id VARCHAR(128) NOT NULL,
    user_metadata TEXT,
    PRIMARY KEY (user_id)
);
```

Issue #239, opened 2021-05-05
Support user metadata


tobiasdiez:
## 🚀 Feature

It would be nice if one could store/receive additional metadata associated to a user. This could be for example some external user ids (e.g. stripe customerId) or additional infos such as the full name etc. This is especially helpful if one uses supertokens as a complete user authentication solution and doesn't have an additional database.

## Implementation details

As a reference, here is the documentation for auth0: https://auth0.com/docs/users/metadata

rishabhpoddar:
@tobiasdiez that is a good idea. Though we would add this as a separate recipe that can be added to an existing login recipe so that users who do not need this, won't get this feature. This would enable us to keep providing the minimum required db schema from our side.

rishabhpoddar:
This feature has been released: https://supertokens.com/docs/thirdpartyemailpassword/common-customizations/usermetadata/about

PR #383, opened 2022-03-04
feat(user_metadata): add support for user metadata

porcellus:
## Summary of change
Adds user metadata support

## Related issues
- #381 
- #239 

## Test Plan

Added multiple tests

## Documentation changes

Most of it will be done as part of the backend implementation.

   - In `build.gradle`

- [ ] Update docs w/ database setup


src/test/java/io/supertokens/test/userMetadata/UserMetadataTest.java
```diff
@@ -0,0 +1,173 @@
...
+        assertNotNull(process.checkOrWaitForEvent(ProcessState.PROCESS_STATE.STOPPED));
+    }
+
+    /**
+     * updates the metadata of a user and merges the result
+     *
+     * @throws Exception
+     */
+    @Test
+    public void updateMetadata() throws Exception {
```
rishabhpoddar, line 120: You should also add a test to test concurrent updates to a user's metadata to ensure race conditions are handled correctly.

src/main/java/io/supertokens/inmemorydb/queries/UserMetadataQueries.java
```diff
@@ -0,0 +1,91 @@
...
+
+        return update(start, QUERY.toString(), pst -> pst.setString(1, userId));
+    }
+
+    public static int setUserMetadata_Transaction(Start start, Connection con, String userId, JsonObject metadata)
+            throws SQLException, StorageQueryException {
+
+        String QUERY = "INSERT INTO " + getConfig(start).getUserMetadataTable()
+                + "(user_id, user_metadata) VALUES(?, ?) "
+                + "ON CONFLICT(user_id) DO UPDATE SET user_metadata=excluded.user_metadata;";
```
rishabhpoddar, line 58: What does `user_metadata=excluded.user_metadata` mean?

src/main/java/io/supertokens/inmemorydb/queries/UserMetadataQueries.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    public static JsonObject getUserMetadata(Start start, String userId) throws SQLException, StorageQueryException {
+        String QUERY = "SELECT user_metadata FROM " + getConfig(start).getUserMetadataTable() + " WHERE user_id = ?";
+        return execute(start, QUERY, pst -> pst.setString(1, userId), result -> {
+            if (result.next()) {
+                JsonParser jp = new JsonParser();
+                return jp.parse(result.getString("user_metadata")).getAsJsonObject();
+            }
+            return null;
```
rishabhpoddar, line 88: I think you should return an empty JSON object here instead of checking for `null` in the core, since maybe in the future we might have to query this in other places in the core too, and at that time, we will have to remember to do null checks.

src/main/java/io/supertokens/inmemorydb/queries/UserMetadataQueries.java
```diff
@@ -0,0 +1,91 @@
...
+        ((ConnectionWithLocks) con).lock(userId + getConfig(start).getUserMetadataTable());
+
+        String QUERY = "SELECT user_metadata FROM " + getConfig(start).getUserMetadataTable() + " WHERE user_id = ?";
+
+        return execute(con, QUERY, pst -> pst.setString(1, userId), result -> {
+            if (result.next()) {
+                JsonParser jp = new JsonParser();
+                return jp.parse(result.getString("user_metadata")).getAsJsonObject();
+            }
+            return null;
```
rishabhpoddar, line 77: I think you should return an empty JSON object here instead of checking for `null` in the core, since maybe in the future we might have to query this in other places in the core too, and at that time, we will have to remember to do null checks.

src/main/java/io/supertokens/storageLayer/StorageLayer.java
```diff
@@ -169,6 +171,13 @@ public static JWTRecipeStorage getJWTRecipeStorage(Main main) {
         return (JWTRecipeStorage) getInstance(main).storage;
     }
 
+    public static UserMetadataSQLStorage getUserMetadataStorage(Main main) {
+        if (getInstance(main) == null) {
+            throw new QuitProgramException("please call init() before calling getStorageLayer");
+        }
+        return (UserMetadataSQLStorage) getInstance(main).storage;
```
rishabhpoddar, line 178: You also need to add:
```java
if (getInstance(main).storage.getType() != STORAGE_TYPE.SQL) {
            // we only support SQL for now
            throw new UnsupportedOperationException("");
        }
```

src/main/java/io/supertokens/usermetadata/UserMetadata.java
```diff
@@ -0,0 +1,66 @@
...
+        return storage.startTransaction((con) -> {
+            JsonObject originalMetadata = storage.getUserMetadata_Transaction(con, userId);
+
+            final JsonObject updatedMetadata = originalMetadata == null ? new JsonObject() : originalMetadata;
+            metadataUpdate.entrySet().forEach((entry) -> {
+                updatedMetadata.remove(entry.getKey());
+                if (!entry.getValue().isJsonNull()) {
+                    updatedMetadata.add(entry.getKey(), entry.getValue());
+                }
+            });
```
rishabhpoddar, line 43: Please extract this into a utils function cause other recipes (like user roles which contains metadata too) will need to use this.

coreDriverInterfaceSupported.json
```diff
@@ -6,6 +6,7 @@
     "2.9",
     "2.10",
     "2.11",
-    "2.12"
+    "2.12",
+    "2.13"
```
rishabhpoddar, line 10: You also need to update `pluginInterfaceSupported.json`. It is checked in the comment for this PR, but the file change here is not there.

src/test/java/io/supertokens/test/userMetadata/UserMetadataTest.java
```diff
@@ -0,0 +1,173 @@
...
+        update.add("cleared", JsonNull.INSTANCE);
+
+        JsonObject updateResult = UserMetadata.updateUserMetadata(process.getProcess(), userId, update);
+
+        JsonObject newMetadata = UserMetadata.getUserMetadata(process.getProcess(), userId);
+
+        assertEquals(updateResult, newMetadata);
+
+        // We removed what we set to null
+        assert (!newMetadata.has("toClear"));
```
rishabhpoddar, line 156:
```suggestion
        assert (!newMetadata.has("cleared"));
```

src/test/java/io/supertokens/test/userMetadata/UserMetadataTest.java
```diff
@@ -0,0 +1,173 @@
...
+
+        // First we create the original
+        UserMetadata.updateUserMetadata(process.getProcess(), userId, originalMetadata);
+
+        JsonObject update = new JsonObject();
+        JsonObject updateSubObject = new JsonObject();
+        updateSubObject.addProperty("subsubupdate", "subnew!");
+        update.add("testUpdate", updateSubObject);
+        update.addProperty("testNew", "new!");
+        update.add("cleared", JsonNull.INSTANCE);
```
rishabhpoddar, line 147: You shold also try by just doing `update.add("cleared", null);`?

rishabhpoddar (changes requested):

src/main/java/io/supertokens/inmemorydb/queries/UserMetadataQueries.java
porcellus: I think returning null from the DB could be useful in the future (e.g.: if we decide the endpoint should return if this was the first time creating or if it was an update). Also, other places in the core could just query this through the recipe function instead of directly accessing the DB. I figured why throw this info away on the interface level.

src/main/java/io/supertokens/inmemorydb/queries/UserMetadataQueries.java
porcellus: it just sets the user_metadata to what would've been inserted if the conflict wasn't there

src/test/java/io/supertokens/test/userMetadata/UserMetadataTest.java
porcellus: That does the same thing in effect: the `add` function translates the null into a JsonNull.

src/main/java/io/supertokens/inmemorydb/queries/UserMetadataQueries.java
rishabhpoddar: Cool

src/main/java/io/supertokens/usermetadata/UserMetadata.java
```diff
@@ -34,13 +34,8 @@ public static JsonObject updateUserMetadata(Main main, @Nonnull String userId, @
...
 
-            final JsonObject updatedMetadata = originalMetadata == null ? new JsonObject() : originalMetadata;
-            metadataUpdate.entrySet().forEach((entry) -> {
-                updatedMetadata.remove(entry.getKey());
-                if (!entry.getValue().isJsonNull()) {
-                    updatedMetadata.add(entry.getKey(), entry.getValue());
-                }
-            });
+            JsonObject updatedMetadata = originalMetadata == null ? new JsonObject() : originalMetadata;
+            MetadataUtils.shallowMergeMetadataUpdate(updatedMetadata, metadataUpdate);
```
rishabhpoddar, line 38: You may want to extrac this function to a global utils file. Since another recipe would be using it. So it would be weird if that recipe will refer directly to Metadata recipe.

src/test/java/io/supertokens/test/userMetadata/api/SetUserMetadataAPITest2_13.java
```diff
@@ -0,0 +1,259 @@
...
+        assertEquals(2, resp.entrySet().size());
+        assertEquals("OK", resp.get("status").getAsString());
+        assert (resp.has("metadata"));
+        JsonObject updateResult = resp.getAsJsonObject("metadata");
+
+        JsonObject newMetadata = UserMetadata.getUserMetadata(process.getProcess(), userId);
+        assertEquals(updateResult, newMetadata);
+
+        // We removed what we set to null
+        assert (!newMetadata.has("toClear"));
```
rishabhpoddar, line 142:
```suggestion
        assert (!newMetadata.has("cleared"));
```

rishabhpoddar: Please make sure this type of error is fixed in all tests.

rishabhpoddar (changes requested):

src/main/java/io/supertokens/usermetadata/UserMetadata.java
porcellus: I've moved the entire `MetadataUtils` into the `Utils` folder. It's still global this way, but I felt like the `Utils` class was getting crowded.

rishabhpoddar: ok

rishabhpoddar (approved):

src/main/java/io/supertokens/usermetadata/UserMetadata.java
```diff
@@ -0,0 +1,62 @@
...
+            throws StorageQueryException, StorageTransactionLogicException {
+        UserMetadataSQLStorage storage = StorageLayer.getUserMetadataStorage(main);
+
+        return storage.startTransaction((con) -> {
+            JsonObject originalMetadata = storage.getUserMetadata_Transaction(con, userId);
+
+            JsonObject updatedMetadata = originalMetadata == null ? new JsonObject() : originalMetadata;
+            MetadataUtils.shallowMergeMetadataUpdate(updatedMetadata, metadataUpdate);
+
+            storage.setUserMetadata_Transaction(con, userId, updatedMetadata);
```
rishabhpoddar, line 41: Add explicit testing for if the select statement returns an empty result and by the time it comes here, another thread has already inserted the row. Then what happens here? Does it retry?

rishabhpoddar (changes requested):

src/main/java/io/supertokens/usermetadata/UserMetadata.java
porcellus: Added the test in the Postgres and MySQL PRs.

rishabhpoddar: okay

rishabhpoddar (approved):

rishabhpoddar:
@porcellus please resolve conflicts

Issue #215, opened 2021-04-13
Using argon instead of bcrypt for password hashing.


rishabhpoddar:
More information here: https://news.ycombinator.com/item?id=15646743#:~:text=forum%20writte...-,I'd%20recommend%20you%20to%20use%20Argon2%20instead%20of%20bcrypt,does%20not%20have%20insecure%20parameters.

rishabhpoddar:
We added argon2 hashing in v3.12. It is not the default however, because in order to use it, one must configure the right settings for their hardware.

ejabu:
👍🏻 Recently I also got a problem when my coworkers using mac, and I'm using Linux.

I found the password_hash value filled in differently

PR #398, opened 2022-03-21
feat: Argon2 hashing

rishabhpoddar:
## Summary of change
- Adds argon2 hashing which is compatible with older versions of a password if they are stored as bcrypt
- Adds new config in yaml file to allow users to decide which password hashing alg to use and to configure params for argon hashing too.

The default setting for argon is taken from [OWASP cheat sheet](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html). Actually, our default config is more than what is in the cheat sheet.

As per the chosen salt and hash length, we have chosen 32 and 64 respectively which is twice of the default value in the java library (16 and 32).

Edit:
- I changed the default config to be what the OWASP cheat sheet said cause we want to be able to a certain number of parallel logins per core by default.

## Related issues
- https://github.com/supertokens/supertokens-core/issues/215

## Test Plan
- [x] Passoword hashing / verification with bcrypt
- [x] Passoword hashing / verification with argon
- [x] Pass invalid password to argon and check that it fails verification
- [x] Pass invalid password to bcrypt and check that it fails verification
- [x] Create password with bcrypt -> change config to use argon -> calling verify function and make sure that it checks using bcrypt
- [x] Create password with argon -> change config to use bcrypt -> calling verify function and make sure that it checks using argon
- [x] Check that default config is argon and its settings too.
- [x] Invalid config error (wrong password hashing alg; negative numbers as argon configs)
- [x] Create password with one set of configs for argon, change those configs and verify existing password
- [x] Change argon config and check that creation of pass takes that into account
- [x] Create password with bcrypt -> change to argon in config -> reset password flow -> verify password -> see it uses argon (and vice versa)
- [x] Create password with bcrypt -> change to argon in config -> update password manually -> verify password -> see it uses argon (and vice versa)
- [x] Same password yields different hash + salt with argon
- [x] (Manual test) -> parallel sign up sign in to see memory consumption

## Documentation changes
TODO

   - In `build.gradle`
- [ ] If there are new dependencies that have been added in `build.gradle`, please make sure to add them in `implementationDependencies.json`.

- [x] Tests
- [x] Confirm that the default argon2 config is ok. And should we allow the user to change the salt and hash length? 
- [x] Should we use argon2i or argon 2id? [OWASP cheat sheet](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html) recommends argon2id, but i read a few blogs saying for password hashing, argon2i is preferred. 
- [x] See how to determine if password hash is done via argon (without using string comparison?)
- [x] Expose argon setting config values
- [x] Add a queue for password hashing to prevent excessive memory usage
- [x] Consider not making argon2 the default given the number of configs that need to be given by the user.
- [x] Allow change to Bcrypt rounds via user config.
- [x] Determine default values for the core (and change the CLI messaging + tests accordingly). We can do this by taking into account a t2 micro instance in AWS as a benchmark.
- [x] Test out in docker image


config.yaml
```diff
@@ -83,4 +83,16 @@ core_config_version: 0
...
 
 # (OPTIONAL | Default: ""). Used to prepend a base path to all APIs when querying the core.
-# base_path:
\ No newline at end of file
+# base_path:
+
+# (OPTIONAL | Default: "ARGON2"). The password hashing algorithm to use. Values are "ARGON2" | "BCRYPT"
+# password_hashing_alg:
+
+# (OPTIONAL | Default: 3). Number of iterations for argon2 password hashing
```
porcellus, line 91: I think these defaults look fine, this is the recommendation for memory-constrained envs. (Check out the [RFC](https://datatracker.ietf.org/doc/rfc9106/), Section 7.4)
We could also test (at least manually) how many parallel login attempts we can handle on an example PC (or AWS instance).

devConfig.yaml
```diff
@@ -83,4 +83,16 @@ core_config_version: 0
 disable_telemetry: true
 
 # (OPTIONAL | Default: ""). Used to prepend a base path to all APIs when querying the core.
-# base_path:
\ No newline at end of file
+# base_path:
+
+# (OPTIONAL | Default: "ARGON2"). The password hashing algorithm to use. Values are "ARGON2" | "BCRYPT"
```
porcellus, line 88: Is the new default intentional? It's not mentioned in the changelog.

config.yaml
porcellus: I think a periodically updated list about this and/or how much time each login takes with some recommended settings could be nice in the docs.
Also, check out: https://www.ory.sh/choose-recommended-argon2-parameters-password-hashing/

porcellus (changes requested):

config.yaml
rishabhpoddar: > I think a periodically updated list about this and/or how much time each login takes with some recommended settings could be nice in the docs.

Agreed.

rishabhpoddar: > I think a periodically updated list about this and/or how much time each login takes with some recommended settings could be nice in the docs.

When you say "periodically updated list", what do you mean?

devConfig.yaml
rishabhpoddar: Yes. Changelog is now updated.

config.yaml
rishabhpoddar: I have changed the default configs. Please see once again

config.yaml
```diff
@@ -83,4 +83,23 @@ core_config_version: 0
...
 
 # (OPTIONAL | Default: ""). Used to prepend a base path to all APIs when querying the core.
-# base_path:
\ No newline at end of file
+# base_path:
+
+# (OPTIONAL | Default: "BCRYPT"). The password hashing algorithm to use. Values are "ARGON2" | "BCRYPT"
+# password_hashing_alg:
+
+# (OPTIONAL | Default: 11). Number of rounds to set for bcrypt password hashing
```
porcellus, line 91: I think this should be `bcrypt_log_rounds` or `bcrypt_cost`.

src/main/java/io/supertokens/config/CoreConfig.java
```diff
@@ -98,6 +117,43 @@ public String getBasePath() {
         return base_path;
     }
 
+    public enum PASSWORD_HASHING_ALG {
+        ARGON2, BCRYPT
+    }
+
+    public int getArgon2HashingPoolSize() {
```
porcellus, line 124: Why do we create the `blockedqueue` in this case? It's fine either way, it just seems unnecessary.

porcellus (changes requested):

config.yaml
rishabhpoddar: sounds good.

src/main/java/io/supertokens/config/CoreConfig.java
rishabhpoddar: Cause it's the simplest way I found to achieve the following:
- We want to run a max of X number of hashes in parallel whilst keeping other hashing requests in a queue.
- A blockedqueue enables this exact use case. When we call `this.boundedQueue.put`, it will block the calling thread if there is no empty space available in the queue (which is of size X).
- It will then unblock the thread as soon as there is space for it in the queue.
- We use the `waiter` object to fill in space in the queue and this object is unique for each thread.

config.yaml
rishabhpoddar: I have updated this to `bcrypt_log_rounds`.

src/main/java/io/supertokens/config/CoreConfig.java
porcellus: I agree. What I meant to ask was why do we normalize the pool size here and create the queue even if it's not used.

porcellus (approved):
LGTM.

src/main/java/io/supertokens/config/CoreConfig.java
rishabhpoddar: Ahh. I had first made the queue so that it was null by default and gets initiated only if argon2 is being used, but some fo the tests change the password hashing alg being used without restarting the core in which case I would get a null pointer exception. 

Sure, I could have also initiated the queue in those tests, but initialising it all the time is fine too (it doesn't take up any memory upfront anyway). So for cleaner tests, i did it this way.

rishabhpoddar:
Error if out of memory:
```
/usr/bin/supertokens: line 9:    15 Killed                  "${ST_INSTALL_LOC}"jre/bin/java -classpath "${ST_INSTALL_LOC}cli/*" io.supertokens.cli.Main false "${ST_INSTALL_LOC}" $@
```

rishabhpoddar:
Argon2 hashing settings for SaaS prod instances:
```
====Final values====
Average time per hash is: 297 MS

argon2_memory_kb: 87795 (85 MB)
argon2_iterations: 1
argon2_parallelism: 2
argon2_hashing_pool_size: 1

====================
```

PR #414, opened 2022-04-06
feat(userroles): adds user role recipe

jscyo:
## Summary of change
Adds UserRoles Recipe

## Related issues
- https://github.com/supertokens/for-zenhub/issues/98

## Test Plan
(Write your test plan here. If you changed any code, please provide us with clear instructions on how you verified your changes work. Bonus points for screenshots and videos!)

## Documentation changes
(If relevant, please create a PR in our [docs repo](https://github.com/supertokens/docs), or create a checklist here highlighting the necessary changes)

## Checklist for important updates
- [x] Changelog has been updated
    - [x] If there are any db schema changes, mention those changes clearly
- [x] `coreDriverInterfaceSupported.json` file has been updated (if needed)
- [x] `pluginInterfaceSupported.json` file has been updated (if needed)
- [x] Changes to the version if needed
   - In `build.gradle`
- [x] Had installed and ran the pre-commit hook
- [ ] ~~If there are new dependencies that have been added in `build.gradle`, please make sure to add them in `implementationDependencies.json`.~~
- [x] Issue this PR against the latest non released version branch.
   - To know which one it is, run find the latest released tag (`git tag`) in the format `vX.Y.Z`, and then find the latest branch (`git branch --all`) whose `X.Y` is greater than the latest released tag.
   - If no such branch exists, then create one from the latest released branch.

## Remaining TODOs for this PR
- none

PR #426, opened 2022-04-18
feat: adds RemoveUserRoleAPI

jscyo:
## Summary of change
- fixs `deleteRoleForUser_Transaction` query 
- Adds the `RemoveUserRoleAPI` and its tests 

## Related issues
- https://github.com/supertokens/supertokens-core/pull/414

### Storage tests:
- [x] test removing an unknown role from a user
- [x] test removing a role from a user

### Core function tests:
- [x] test removing an unknown role from a user, it should throw a `UNKNOWN_ROLE_EXCEPTION`
- [ ] test removing a role from a user
- [x] assign multiple roles to a user, delete a role, check that the user doesn't have the deleted role but still has the other roles

### API tests:
- [x] badInputTests
  - [x] empty request
  - [x] don't pass userId
  - [x] pass userId as a number
  - [x] don't pass role
  - [ ] pass role as a number
  - [x] pass role an empty string
- [x] add a role to a user, remove the role, check that the user has no role
- [x] remove an unknown role from a user
 
## Documentation changes
(If relevant, please create a PR in our [docs repo](https://github.com/supertokens/docs), or create a checklist here highlighting the necessary changes)

- [ ] ~~Changelog has been updated~~
   - ~~In `build.gradle`~~
- [ ] ~~Issue this PR against the latest non released version branch.~~
   - ~~To know which one it is, run find the latest released tag (`git tag`) in the format `vX.Y.Z`, and then find the latest branch (`git branch --all`) whose `X.Y` is greater than the latest released tag.~~
   - ~~If no such branch exists, then create one from the latest released branch.~~

### DIFF
--- a/src/main/java/io/supertokens/userroles/UserRoles.java
+++ b/src/main/java/io/supertokens/userroles/UserRoles.java
@@ -63,24 +63,30 @@ public class UserRoles {
     }
 
     // remove a role mapped to a user, if the role doesn't exist throw a UNKNOWN_ROLE_EXCEPTION error
-//    public static void deleteUserRole(Main main, String userId, String role)
-//            throws StorageQueryException, StorageTransactionLogicException {
-//
-//        UserRolesSQLStorage storage = StorageLayer.getUserRolesStorage(main);
-//
-//        storage.startTransaction(con -> {
-//
-//            boolean doesRoleExist = storage.doesRoleExist_Transaction(con, role);
-//
-//            if (doesRoleExist) {
-//                storage.deleteRoleForUser_Transaction(con, userId, role);
-//            } else {
-//                throw new UnknownRoleException();
-//            }
-//
-//            return null;
-//        });
-//    }
+    public static void removeUserRole(Main main, String userId, String role)
+            throws StorageQueryException, StorageTransactionLogicException, UnknownRoleException {
+
+        UserRolesSQLStorage storage = StorageLayer.getUserRolesStorage(main);
+
+        try {
+            storage.startTransaction(con -> {
+
+                boolean doesRoleExist = storage.doesRoleExist_Transaction(con, role);
+
+                if (doesRoleExist) {
+                    storage.deleteRoleForUser_Transaction(con, userId, role);
+                } else {
+                    throw new StorageTransactionLogicException(new UnknownRoleException());
+                }
+                return null;
+            });
+        } catch (StorageTransactionLogicException e) {
+            if (e.actualException instanceof UnknownRoleException) {
+                throw (UnknownRoleException) e.actualException;
+            }
+            throw e;
+        }
+    }
 //
 //    // retrieve all roles associated with the user
 //    public static String[] getRolesForUser(Main main, String userId) throws StorageQueryException {
--- a/src/main/java/io/supertokens/userroles/UserRoles.java
+++ b/src/main/java/io/supertokens/userroles/UserRoles.java
@@ -87,7 +87,7 @@ public class UserRoles {
             throw e;
         }
     }
-//
+
 //    // retrieve all roles associated with the user
 //    public static String[] getRolesForUser(Main main, String userId) throws StorageQueryException {
 //        return StorageLayer.getUserRolesStorage(main).getRolesForUser(userId);
--- a/src/main/java/io/supertokens/userroles/UserRoles.java
+++ b/src/main/java/io/supertokens/userroles/UserRoles.java
@@ -65,22 +65,21 @@ public class UserRoles {
     }
 
     // remove a role mapped to a user, if the role doesn't exist throw a UNKNOWN_ROLE_EXCEPTION error
-    public static void removeUserRole(Main main, String userId, String role)
+    public static boolean removeUserRole(Main main, String userId, String role)
             throws StorageQueryException, StorageTransactionLogicException, UnknownRoleException {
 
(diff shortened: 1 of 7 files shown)


src/main/java/io/supertokens/userroles/UserRoles.java
```diff
@@ -63,25 +63,31 @@ public static boolean doesRoleExist(Main main, String role) throws StorageQueryE
...
+                    throw new StorageTransactionLogicException(new UnknownRoleException());
+                }
+                return null;
+            });
+        } catch (StorageTransactionLogicException e) {
+            if (e.actualException instanceof UnknownRoleException) {
+                throw (UnknownRoleException) e.actualException;
+            }
+            throw e;
+        }
```
rishabhpoddar, line 88: You do not need to start a transaction.. you can just call `deleteRoleForUser_Transaction` and if it returns `false`, you can throw `UnknownRoleException`, else just return.

So you can make it just `deleteRoleForUser` (will require change to plugin interface). How are we doing this (or plan to do this) for removing role and permission - discuss on a call

rishabhpoddar (changes requested):

src/test/java/io/supertokens/test/userRoles/UserRolesTest.java
```diff
@@ -301,4 +302,107 @@ public void testAddRoleToUserResponses() throws Exception {
         assertNotNull(process.checkOrWaitForEvent(ProcessState.PROCESS_STATE.STOPPED));
     }
```
rishabhpoddar, line 303: You should add more tests that test race conditions:
- [ ] Deleting a role whilst it's being added to a user -> if the delete succeeds then adding should throw unknown role error (this is a test related to adding role API which we had missed in the previous PR) -> this should be a different PR to the base branch.
- [ ] Deleting a role whilst it's being removed from a user -> if the delete succeeds then deleting role from user should throw unknown role error.
- Any other?
-------------
On a related note, do we even need `doesRoleExist_Transaction` anymore? Where all is this being (or will be) used? Discuss on call.

rishabhpoddar (changes requested):

src/main/java/io/supertokens/userroles/UserRoles.java
jscyo: > You do not need to start a transaction.. you can just call deleteRoleForUser_Transaction and if it returns false, you can throw UnknownRoleException, else just return.

Hey this works for the case where the role doesn't exist, but, if the role exists but it hasn't been mapped to a user, this will still throw an `UNKNOWN_ROLE_EXCEPTION`

src/test/java/io/supertokens/test/userRoles/UserRolesTest.java
jscyo: > On a related note, do we even need doesRoleExist_Transaction anymore? Where all is this being (or will be) used? Discuss on call.
- replied to this in a previous comment

> You should add more tests that test race conditions:
- will add the tests

jscyo: >Deleting a role whilst it's being added to a user -> if the delete succeeds then adding should throw unknown role error (this is a test related to adding role API which we had missed in the previous PR) -> this should be a different PR to the base branch.)

Added this test but with modifications

>  Deleting a role whilst it's being added to a user -> if the delete succeeds then adding should throw unknown role error (this is a test related to adding role API which we had missed in the previous PR) -> this should be a different PR to the base branch.)

This test does not make sense as the query used for this operation will throw a foreign key constraint error when trying to add a role which doesn't exist to a user, As opposed to the other test where a delete query can succeed when the role does not exist

src/test/java/io/supertokens/test/userRoles/UserRolesTest.java
```diff
@@ -319,4 +320,115 @@ public void testAddRoleToUserResponses() throws Exception {
...
+            Utils.checkThatArraysAreEqual(roles, userRoles);
+        }
+
+        {
+            // remove the roles from the user
+            boolean didUserHaveRole = UserRoles.removeUserRole(process.main, userId, roles[0]);
+            assertTrue(didUserHaveRole);
+
+            // check that the user has no roles
+            storage.getRolesForUser(userId);
```
rishabhpoddar, line 378: this is incomplete?

src/test/java/io/supertokens/test/userRoles/api/RemoveUserRoleAPITest.java
```diff
@@ -0,0 +1,283 @@
...
+                HttpRequestForTesting.sendJsonPOSTRequest(process.getProcess(), "",
+                        "http://localhost:3567/recipe/user/role/remove", request, 1000, 1000, null,
+                        Utils.getCdiVersion2_14ForTests(), "userroles");
+                throw new Exception("should not come here");
+            } catch (HttpResponseException e) {
+                System.out.println(e.getMessage());
+                assertTrue(e.statusCode == 400 && e.getMessage().equals(
+                        "Http error. Status Code: 400. Message:" + " Field name 'role' is invalid in JSON input"));
+            }
+        }
```
rishabhpoddar, line 156: this is done twice?

rishabhpoddar (changes requested):

rishabhpoddar (approved):

Issue #382, opened 2022-03-03
Define DB schema and APIs for UserRoles


jscyo:
## 🚀 Feature

Define the DB schema and APIs for user roles and permissions

**NOTE :  The delete user API should also call the remove user roles API**
### DB schema
-----------
#### MySQL
```sql
CREATE TABLE IF NOT EXISTS roles (
  role VARCHAR(255) NOT NULL,
  PRIMARY KEY (role)  
);

CREATE TABLE IF NOT EXISTS role_permissions (
  role VARCHAR(255) NOT NULL,
  permission VARCHAR(255) NOT NULL,
  PRIMARY KEY (role, permission),
  FOREIGN KEY (role) REFERENCES roles(role) ON DELETE CASCADE
);

CREATE INDEX role_permissions_permission_index ON role_permissions(permission);

CREATE TABLE IF NOT EXISTS user_roles (
  user_id VARCHAR(128) NOT NULL,
  role VARCHAR(255) NOT NULL,
  PRIMARY KEY (user_id, role),
  FOREIGN KEY (role)  REFERENCES roles (role) ON DELETE CASCADE
);

CREATE INDEX user_roles_role_index ON user_roles(role);
```

#### PSQL
```sql
CREATE TABLE IF NOT EXISTS roles (
  role VARCHAR(255) NOT NULL,
  CONSTRAINT roles_pkey PRIMARY KEY (role)  
);

CREATE TABLE IF NOT EXISTS role_permissions (
  role VARCHAR(255) NOT NULL,
  permission VARCHAR(255) NOT NULL,
  CONSTRAINT role_permissions_pkey PRIMARY KEY (role, permission),
  CONSTRAINT role_permissions_role_fkey FOREIGN KEY (role) REFERENCES roles(role) ON DELETE CASCADE
);

CREATE INDEX role_permissions_permission_index ON role_permissions(permission);

CREATE TABLE IF NOT EXISTS user_roles (
  user_id VARCHAR(128) NOT NULL,
  role VARCHAR(255) NOT NULL,
  CONSTRAINT user_roles_pkey PRIMARY KEY (user_id, role),
  CONSTRAINT user_roles_role_fkey FOREIGN KEY (role)  REFERENCES roles (role) ON DELETE CASCADE
);

CREATE INDEX user_roles_role_index ON user_roles(role);
```
-------

### Core API spec
[...]

PR #428, opened 2022-04-20
fix: updates plugin-interface functions, test fixs

jscyo:
## Summary of change
The plugin-interface has been updated to account for the discussion mentioned [here](https://github.com/supertokens/supertokens-core/issues/382#issuecomment-1102458937). The changes in this PR are to account for the plugin-interface updates.

#### SQLite changes:
- `createNewRoleOrDoNothingIfExists_Transaction` now returns a boolean, so updated the query to return a boolean on whether a row was updated or not.

#### Core function changes:
- `addRoleToUser` now returns a boolean on whether a role was added to a user or the mapping already existed 
- `createNewRoleOrModifyItsPermissions` now returns a boolean on whether a new role was created

#### API changes:
- `AddRoleToUserAPI` now returns an additional property in its response called `didUserHaveRole`
- `CreateRoleAPI` now returns an additional property in its response called `createdNewRole`

### Test changes:
- Updates original tests to account for these changes
- Adds additional API tests to test the new fields in the responses
-
## Related issues
- https://github.com/supertokens/supertokens-core/issues/382#issuecomment-1102458937

## Test Plan
- Added additional tests

## Documentation changes
- none

- [ ] ~~Changelog has been updated~~
   - ~~In `build.gradle`~~
- [ ] ~~Had installed and ran the pre-commit hook~~
- [ ] ~~Issue this PR against the latest non released version branch.~~
   - ~~To know which one it is, run find the latest released tag (`git tag`) in the format `vX.Y.Z`, and then find the latest branch (`git branch --all`) whose `X.Y` is greater than the latest released tag.~~
   - ~~If no such branch exists, then create one from the latest released branch.~~

- none

### DIFF
--- a/src/main/java/io/supertokens/webserver/api/userroles/AddUserRoleAPI.java
+++ b/src/main/java/io/supertokens/webserver/api/userroles/AddUserRoleAPI.java
@@ -59,10 +59,10 @@ public class AddUserRoleAPI extends WebserverAPI {
         }
 
         try {
-            boolean didUserHaveRole = !UserRoles.addRoleToUser(main, userId, role);
+            boolean didUserAlreadyHaveRole = !UserRoles.addRoleToUser(main, userId, role);
             JsonObject response = new JsonObject();
             response.addProperty("status", "OK");
-            response.addProperty("didUserHaveRole", didUserHaveRole);
+            response.addProperty("didUserAlreadyHaveRole", didUserAlreadyHaveRole);
             super.sendJsonResponse(200, response, resp);
         } catch (UnknownRoleException e) {
             JsonObject response = new JsonObject();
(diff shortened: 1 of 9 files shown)


src/main/java/io/supertokens/webserver/api/userroles/AddUserRoleAPI.java
```diff
@@ -59,9 +59,10 @@ protected void doPut(HttpServletRequest req, HttpServletResponse resp) throws IO
         }
 
         try {
-            UserRoles.addRoleToUser(main, userId, role);
+            boolean didUserHaveRole = !UserRoles.addRoleToUser(main, userId, role);
             JsonObject response = new JsonObject();
             response.addProperty("status", "OK");
+            response.addProperty("didUserHaveRole", didUserHaveRole);
```
rishabhpoddar, line 65: I think a better name would be `didUserAlreadyHaveRole`?

rishabhpoddar (changes requested):

src/main/java/io/supertokens/webserver/api/userroles/AddUserRoleAPI.java
jscyo: agreed will make the change

rishabhpoddar (approved):

PR #429, opened 2022-04-20
feat: Windows Support

JeremyEastham:
## Summary of Change
Trying to contribute to this project has been a challenge to set up on Windows. The purpose of this PR is to streamline the project setup process for other developers that use Windows on their development machine. I have fixed a number of Windows-specific bugs and added extensive Windows development setup instructions to `CONTRIBUTING.md` (for both Git Bash and WSL2).

Before starting work on this PR, **only 87% of tests were passing when run on Windows** via Git Bash. I chose to use Git Bash because it uses the native Windows version of Java at runtime, while still being able to execute the bash scripts in `supertokens-root`. It was important to me that the native Windows version of Java was used so that I could make sure that all core functionality remained intact when the core was run on a Windows machine. Many of the tests were failing due to bugs in `Utils.java`, which did not use `System.lineSeparator()` correctly when mocking config values. Additionally, a few tests compared exception message strings that were not identical cross-platform.

## Related Issues
(no related issues)

## Test Plan
I have removed a portion of `WebserverTest.differentHostNameTest()`. This test was failing because the core was not visible to any of my local IP addresses when hosted on `localhost` due to my firewall settings. Thus, it was not an issue with the code, but instead, an issue with the test. This may be related to how Windows handles `localhost` addresses, but documentation is sparse on this issue.

The purpose of this PR is not to add functionality, so no tests have been added.

## Documentation Changes
(no changes needed)

- [X] Changelog has been updated
    - [X] If there are any db schema changes, mention those changes clearly
- [X] `coreDriverInterfaceSupported.json` file has been updated (if needed)
- [X] `pluginInterfaceSupported.json` file has been updated (if needed)
- [X] Changes to the version if needed
   - In `build.gradle`
- [X] Had installed and ran the pre-commit hook
- [X] If there are new dependencies that have been added in `build.gradle`, please make sure to add them in `implementationDependencies.json`.
- [X] Issue this PR against the latest non released version branch.

- [ ] Figure out why "Running the Core (Manually)" instructions are not working
    - [ ] Rewrite instructions in `CONTRIBUTING.md` if needed
- [ ] Merge supertokens/supertokens-root#8
    - Instructions added to `CONTRIBUTING.md` are dependent on these run configurations existing

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,118 +3,207 @@
 We're so excited you're interested in helping with SuperTokens! We are happy to help you get started, even if you don't have any previous open-source experience :blush:
 
 ## New to Open Source?
-1. Take a look at [How to Contribute to an Open Source Project on GitHub](https://egghead.io/courses/how-to-contribute-to-an-open-source-project-on-github)
-2. Go through the [SuperTokens Code of Conduct](https://github.com/supertokens/supertokens-core/blob/master/CODE_OF_CONDUCT.md)
+1. Take a look at [How to Contribute to an Open Source Project on GitHub][how-to-contribute-to-open-source]
+2. Go through the [SuperTokens Code of Conduct][code-of-conduct]
 
-## Where to ask Questions?
-1. Check our [Github Issues](https://github.com/supertokens/supertokens-core/issues) to see if someone has already answered your question.  
-2. Join our community on [Discord](https://supertokens.io/discord) and feel free to ask us your questions  
+## Where to Ask Questions?
+1. Check our [Github Issues][issues] to see if someone has already answered your question.
+2. Join our community on [Discord][discord] and feel free to ask us your questions
 
-As you gain experience with SuperTokens, please help answer other people's questions! :pray: 
+As you gain experience with SuperTokens, please help answer other people's questions! :pray:
 
 ## What to Work On?
-You can get started by taking a look at our [Github issues](https://github.com/supertokens/supertokens-core/issues)  
-If you find one that looks interesting and no one else is already working on it, comment in the issue that you are going to work on it.  
+You can get started by taking a look at our [Github Issues][issues]  
+If you find one that looks interesting and no one else is already working on it, comment in the issue that you are going to work on it.
 
-Please ask as many questions as you need, either directly in the issue or on [Discord](https://supertokens.io/discord). We're happy to help!:raised_hands:
+Please ask as many questions as you need, either directly in the issue or on [Discord][discord]. We're happy to help! :raised_hands:
 
-### Contributions that are ALWAYS welcome 
-
+### Contributions that are ALWAYS welcome
 1. More tests
-2. Contributing to discussions that can be found [here](https://github.com/supertokens/supertokens-core/issues?q=is%3Aissue+is%3Aopen+label%3Adiscussions)
+2. Contributing to discussions that can be found [here][issues-discussions]
 3. Improved error messages
 4. Educational content like blogs, videos, courses
-
 
 ## Development Setup
 
 ### With Gitpod
-1. Navigate to the [supertokens-root](https://github.com/supertokens/supertokens-root) repository
+1. Navigate to the [supertokens-root][supertokens-root] repository
 2. Click on the `Open in Gitpod` button
 
 ### Local Setup Prerequisites
-- OS: Linux or macOS. Or if using Windows, you need to use [wsl2](https://docs.microsoft.com/en-us/windows/wsl/about).
-- JDK: openjdk 15.0.1. Installation instructions for Mac and Linux can be found in [our wiki](https://github.com/supertokens/supertokens-core/wiki/Installing-OpenJDK-for-Mac-and-Linux)
-- IDE: [IntelliJ](https://www.jetbrains.com/idea/download/)(recommended) or equivalent IDE
+- OS: Linux or macOS. Windows is experimentally supported with Git Bash (recommended, installed with [Git for Windows][git-for-windows]) or [WSL2][wsl-about].
+- JDK: OpenJDK 15.0.1
+    - Installation instructions for Mac and Linux can be found in [our wiki][wiki-openjdk-install-instructions-mac-linux]
+    - To install for Windows, download and run the correct installer from the [AdoptOpenJDK Archive][adopt-openjdk-archive]
+- IDE: [IntelliJ IDEA][intellij] (recommended) or equivalent IDE
 
-### Familiarize yourself with SuperTokens
-1. [Architecture of SuperTokens](https://github.com/supertokens/supertokens-core/wiki/SuperTokens-Architecture)
-2. [SuperTokens code and file structure overview](https://github.com/supertokens/supertokens-core/wiki/Code-and-file-structure-overview)
-3. [Versioning methodology](https://github.com/supertokens/supertokens-core/wiki/Versioning,-git-and-releases)
+### Familiarize Yourself with SuperTokens
+1. [Architecture of SuperTokens][wiki-architecture]
+2. [SuperTokens Code and File Structure Overview][wiki-overview]
+3. [Versioning methodology][wiki-versioning]
 
+### Experimental Windows Setup ([Git Bash][git-for-windows]) [Recommended]
+1. Git Bash is automatically installed with [Git for Windows][git-for-windows]
+2. Install the correct version of OpenJDK (see above) from [AdoptOpenJDK][adopt-openjdk-archive]
+3. Open Command Prompt or PowerShell and run `sh` to open Git Bash
+4. Run `echo "TERM=cygwin" >> .bashrc`
+    - This ensures that the correct encoding is used in the console, especially when running `gradle` commands
+5. Run `git config --global core.autocrlf false`
+    - This ensures that line endings are not converted to Windows (CRLF) line endings on checkout
+    - Scripts will fail if line endings are CRLF
+6. Follow the Project Setup below (using Git Bash)
+7. If using IntelliJ IDEA:
+    - Go to `File > Project Structure... > Project Settings > Project` and set `Project SDK` to the correct OpenJDK version
+    - Go to `File > Settings... > Tools > Terminal` and set `Application Settings > Shell Path` to the location of Git Bash
+        - For the default install directory this is `C:\Program Files\Git\bin\bash.exe`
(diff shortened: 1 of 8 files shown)


rishabhpoddar:
@JeremyEastham please also make sure to run the tests with your modifications via github actions so that the failing PR check passes.

JeremyEastham:
Tests are failing now because I renamed `./startTestingEnv` to `./startTestEnv` in supertokens/supertokens-root#9. Once that PR is merged, tests *should* pass. The problem is that I changed the call in *this* PR to `./startTestEnv`, which does not exist in supertokens-root master yet.

rishabhpoddar (changes requested):

src/test/java/io/supertokens/test/Utils.java
```diff
@@ -31,22 +32,21 @@
 
 import java.io.*;
 import java.nio.charset.StandardCharsets;
```
rishabhpoddar, line 34: Changes to this file should also be reflected in the plugin repos for the core.

rishabhpoddar (changes requested):

src/test/java/io/supertokens/test/AuthRecipeTest.java
```diff
@@ -205,8 +205,11 @@ public void paginationTest() throws Exception {
         }
 
         UserInfo user1 = EmailPassword.signUp(process.getProcess(), "test0@example.com", "password0");
+        Thread.sleep(1); // Sleep to remove race condition
```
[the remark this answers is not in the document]
JeremyEastham: One of the times that I ran this test, the test failed because when the users were sorted by creation time, they were not sorted in the correct order. This is due to Java's implementation of `System.currentTimeMillis()` that does not necessarily return a unique timestamp on each call.

See this excerpt from the Java documentation:
> `public static long currentTimeMillis()`
> 
> Returns the current time in milliseconds. Note that while the unit of time of the return value is a millisecond, the granularity of the value depends on the underlying operating system and may be larger. For example, many operating systems measure time in units of tens of milliseconds.

For Example:
```java
for(int i = 0; i < 50; i++)
    System.out.println(System.currentTimeMillis());
```
When I ran this code on my Windows machine, it printed out `1650942035022` many times, followed by `1650942035026` many more times.

However, if I run this second example:
```java
for(int i = 0; i < 50; i++) {
    System.out.println(System.currentTimeMillis());
    Thread.sleep(1);
}
```
The above prints out 50 unique timestamps, 1-3 milliseconds apart. This is due to Java's implementation of `Thread.sleep()`, which is guaranteed to sleep *at least* the number of milliseconds that it is given, dependent on the resolution of the clock of the underlying operating system.

src/test/java/io/supertokens/test/WebserverTest.java
```diff
@@ -597,26 +595,12 @@ public void differentHostNameTest() throws InterruptedException, IOException, Ht
...
         Utils.setValueInConfig("host", "\"127.0.0.1\"");
         hello("localhost", "3567");
         hello("127.0.0.1", "3567");
-        try {
-            hello(inetAddress.getHostAddress(), "3567");
-            if (!inetAddress.getHostAddress().equals("127.0.0.1")) {
-                fail();
-            }
-        } catch (ConnectException ignored) {
-        }
```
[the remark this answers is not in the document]
JeremyEastham: This test was failing because the core was not visible to any of my local IP addresses when hosted on localhost due to my firewall settings. This may also be related to how Windows handles localhost addresses, but documentation is sparse on this issue.

src/test/java/io/supertokens/test/Utils.java
JeremyEastham: Although I changed the implementations of these methods to make them more readable, the method signatures did not change. No changes should be needed in the plugin repos.

src/test/java/io/supertokens/test/WebserverTest.java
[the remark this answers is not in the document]
rishabhpoddar: I see. I would require these tests to be there.. Maybe we can change the code here to not run if the OS is windows?

src/test/java/io/supertokens/test/Utils.java
```diff
@@ -138,46 +137,26 @@ public static void reset() {
...
-                originalFileContent.append(currentReadingLine).append(System.lineSeparator());
-                currentReadingLine = reader.readLine();
-            }
-            String modifiedFileContent = originalFileContent.toString().replaceAll(oldStr, newStr);
-            try (BufferedWriter writer = new BufferedWriter(new FileWriter("../config.yaml"))) {
-                writer.write(modifiedFileContent);
-            }
-        }
+        String find = "\r?\n#?\\s*" + Pattern.quote(key) + ":.*\r?\n";
+        String replace = newLine + key + ": " + value + newLine;
```
rishabhpoddar, line 158: Do we need to change the regex expression? The older one worked just fine. Did the older one cause issues in windows?

CONTRIBUTING.md
```diff
@@ -3,118 +3,232 @@
...
 5. Set the config variables in the drop down:
-    - **supertokens-plugin-interface repo owner name**: If you have forked the supertokens-plugin-interface repo, then set the value of this to your github username.
-    - **supertokens-plugin-interface repos branch name**: If the core version you are working on is compatible with a plugin-interface version that is not in the master branch, then set the correct branch name in this value.
-6. Click on "Run workflow".
+    - **supertokens-plugin-interface repo owner name**: If you have forked the supertokens-plugin-interface repo, then set the value of this to your GitHub username
+    - **supertokens-plugin-interface repos branch name**: If the core version you are working on is compatible with a plugin-interface version that is not in the master branch, then set the correct branch name in this value
+6. Click on "Run workflow"
 
-## Running the core manually
+## Running the Core (Manually)
```
rishabhpoddar, line 162:
```suggestion
## Running the Core (Using Java)

```

CONTRIBUTING.md
```diff
@@ -3,118 +3,232 @@
...
 1. Run `startTestingEnv --wait` in a terminal, and keep it running
```
rishabhpoddar, line 163:
```suggestion
1. Run `startTestEnv --wait` in a terminal, and keep it running

```

CONTRIBUTING.md
```diff
@@ -3,118 +3,232 @@
...
 2. Then open `supertokens-root` in another terminal and run `cp ./temp/config.yaml .`
-3. Then run `java -classpath "./core/*:./plugin-interface/*" io.supertokens.Main ./ DEV`. This will start the core to listen on `http://localhost:3567`
+3. Then run `java -classpath "./supertokens-core/*:./supertokens-plugin-interface/*" io.supertokens.Main ./ DEV`. This will start the core to listen on `http://localhost:3567`
```
rishabhpoddar, line 165: Are you sure this change is right?

CONTRIBUTING.md
```diff
@@ -3,118 +3,232 @@
...
```
rishabhpoddar, line 166: We should also have instrs on running the core via the `./runCore` script.

rishabhpoddar (changes requested):

CONTRIBUTING.md
JeremyEastham: This is incorrect. I wrote this before I wrote the `runCore` script. I'm going to replace these instructions with instructions to run `runCore`.

CONTRIBUTING.md
JeremyEastham: See above.

src/test/java/io/supertokens/test/Utils.java
JeremyEastham: Yes, this change is necessary.

```
Old Regex: "\n((#\\s)?)" + key                + "(:|((:\\s).+))\n"
New Regex: "\r?\n#?\\s*" + Pattern.quote(key) + ":.*\r?\n"
```

Issues with the Old Regex:
- It used `\n` instead of `\r?\n`, which did not match Windows line endings (`\r\n`)
- It only matched comments with a single whitespace character after it
- It did not match non-commented keys with leading whitespace
- It was also much harder to read due to unnecessary group nesting and the unnecessary `|` operator
- It did not quote the key, so any regex characters in the key would make it fail

When reviewing this, I actually noticed one issue with the New Regex. It did not match commented lines with leading whitespace. I have committed this change.

CONTRIBUTING.md
JeremyEastham: I have clarified this in this document. It now reads `With Console`.

CONTRIBUTING.md
JeremyEastham: This part of `CONTRIBUTING.md` has been replaced by instructions to run `runCore`.

src/test/java/io/supertokens/test/Utils.java
rishabhpoddar: Ok. So this change would be required in the plugins repos as well. Since they too have a similar function in there.

src/test/java/io/supertokens/test/WebserverTest.java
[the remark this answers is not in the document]
JeremyEastham: I thought about that, but wouldn't the situation be similar on other operating systems as well? If settings are strict (as they probably should be), access to localhost servers would not be visible to LAN addresses.

src/test/java/io/supertokens/test/Utils.java
JeremyEastham: Okay, I see. I was thinking that the plugin repos called `io.supertokens.test.Utils.commentConfigValue()` and `io.supertokens.test.Utils.setValueInConfig()` directly.

I was also thinking about renaming `setValueInConfig()` to `setConfigValue()` to match `commentConfigValue()`. If I rename all references in `supertokens-core`, will any change be needed in other repositories?

src/test/java/io/supertokens/test/WebserverTest.java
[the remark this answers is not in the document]
rishabhpoddar: It has more or less worked with all linux based systems, mac, CICD, GitHub action and everyone in our dev team.. so i don't think it's too much of a problem. I just want this to also run in CICD / GH action pipeline. It would be OK to have them skipped in local dev env.

src/test/java/io/supertokens/test/Utils.java
rishabhpoddar: You should change them in the plugin repos too, just to be consistent.

Issue #499, opened 2022-08-31
access token payload property is omitted when value is 'null'


n1ru4l:
## 🐛 Bug Report

I noticed that when I decode the access token payload a `null` value property is actually omitted. I expected it to keep the value `null` within the payload.

At the same time within the `sessionData` the null value exists in the payload (within the  database).

```ts
const SuperTokenAccessTokenModel = zod.object({
  version: zod.literal('1'),
  superTokensUserId: zod.string(),
  /**
   * Supertokens for some reason omits externalUserId from the access token payload if it is null.
   */
  externalUserId: zod.optional(zod.union([zod.string(), zod.null()])),
  email: zod.string(),
});
```

## Useful informations

```ts
SessionNode.init({
  override: {
    functions: originalImplementation => {
      return {
        ...originalImplementation,
        createNewSession: async function (input) {
          const user = await ThirdPartyEmailPasswordNode.getUserById(input.userId);

          if (!user) {
            throw new Error(
              `SuperTokens: Creating a new session failed. Could not find user with id ${input.userId}.`
            );
          }

          const externalUserId = user.thirdParty ? `${user.thirdParty.id}|${user.thirdParty.userId}` : null;

          input.accessTokenPayload = {
            version: '1',
            superTokensUserId: input.userId,
            externalUserId, // This is potentially null
            email: user.email,
          };

          input.sessionData = {
            version: '1',
            superTokensUserId: input.userId,
            externalUserId,
            email: user.email,
          };

          return originalImplementation.createNewSession(input);
        },
      };
    },
  },
})
```

porcellus:
Hi, thanks for reporting this. This could be a bug in our current version.
However, we are planning to move to "merging" access token payloads instead of replacing the entire object in the near future.
Does your use case require keeping a null property in the root of the access token payload, or would you be OK with putting your user info into an object inside the payload?

n1ru4l:
> Does your use case require keeping a null property in the root of the access token payload, or would you be OK with putting your user info into an object inside the payload?

I can also put it out of the root, but coming mostly form a JavaScript background I did not expected a `null` value being dropped.

E.g. if you call `JSON.stringify` you get the following results:

`JSON.stringify({ foo: null })` -> `'{"foo":null}'`
`JSON.stringify({ foo : undefined })` -> `'{}'`

> However, we are planning to move to "merging" access token payloads instead of replacing the entire object in the near future.

[...]

PR #500, opened 2022-09-01
fix: null props in access token

porcellus:
## Summary of change
Fixes handling of nulls in access token payload

## Related issues
- #499 

## Test Plan

- Added a separate test in session token creation
- Updated other tests that check if access token payload is kept consistent

## Documentation changes
N/A, bugfix

   - In `build.gradle`

### DIFF
--- a/src/main/java/io/supertokens/session/accessToken/AccessToken.java
+++ b/src/main/java/io/supertokens/session/accessToken/AccessToken.java
@@ -146,7 +146,8 @@ public class AccessToken {
         }
         AccessTokenInfo accessToken = new AccessTokenInfo(sessionHandle, userId, refreshTokenHash1, expiryTime,
                 parentRefreshTokenHash1, userData, antiCsrfToken, now, lmrt);
-        String token = JWT.createJWT(new Gson().toJsonTree(accessToken), signingKey.privateKey, VERSION.V2);
+
+        String token = JWT.createJWT(Utils.toJsonTreeWithNulls(accessToken), signingKey.privateKey, VERSION.V2);
         return new TokenInfo(token, expiryTime, now);
 
     }
@@ -166,7 +167,7 @@ public class AccessToken {
         accessToken = new AccessTokenInfo(sessionHandle, userId, refreshTokenHash1, expiryTime, parentRefreshTokenHash1,
                 userData, antiCsrfToken, now, null);
 
-        String token = JWT.createJWT(new Gson().toJsonTree(accessToken), signingKey.privateKey, VERSION.V1);
+        String token = JWT.createJWT(Utils.toJsonTreeWithNulls(accessToken), signingKey.privateKey, VERSION.V1);
         return new TokenInfo(token, expiryTime, now);
 
     }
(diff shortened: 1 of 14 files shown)


src/main/java/io/supertokens/session/info/SessionInformationHolder.java
```diff
@@ -44,4 +49,11 @@ public SessionInformationHolder(@Nonnull SessionInfo session, @Nullable TokenInf
         this.idRefreshToken = idRefreshToken;
         this.antiCsrfToken = antiCsrfToken;
     }
+
+    public JsonObject toJsonObject() {
+        JsonObject json = new Gson().toJsonTree(this).getAsJsonObject();
+        json.add("session", Utils.toJsonTreeWithNulls(session));
```
rishabhpoddar, line 55: are you sure this replaces the existing "session" key in the object and is not ignored?

src/test/java/io/supertokens/test/session/api/RefreshSessionAPITest2_7.java
```diff
@@ -363,6 +364,7 @@ public void successOutputWithValidRefreshTokenTest() throws Exception {
 
         String userId = "userId";
         JsonObject userDataInJWT = new JsonObject();
+        userDataInJWT.add("nullProp", JsonNull.INSTANCE);
```
rishabhpoddar, line 367: you should check that post refresh (in this test), that this still has the `nullProp` key / value

src/main/java/io/supertokens/session/accessToken/AccessToken.java
```diff
@@ -146,7 +146,8 @@ public static TokenInfo createNewAccessToken(@Nonnull Main main, @Nonnull String
         }
         AccessTokenInfo accessToken = new AccessTokenInfo(sessionHandle, userId, refreshTokenHash1, expiryTime,
                 parentRefreshTokenHash1, userData, antiCsrfToken, now, lmrt);
-        String token = JWT.createJWT(new Gson().toJsonTree(accessToken), signingKey.privateKey, VERSION.V2);
+
+        String token = JWT.createJWT(Utils.toJsonTreeWithNulls(accessToken), signingKey.privateKey, VERSION.V2);
         return new TokenInfo(token, expiryTime, now);
```
rishabhpoddar, line 152: - Would the JWT recipe also need to change to allow setting for null values similar to how you have changed the session recipe? Please add a test for it as well in the core.
- What about session data (thats stored in the db)

rishabhpoddar (changes requested):

src/main/java/io/supertokens/session/accessToken/AccessToken.java
porcellus: - The JWT library we use throws null claims away (https://github.com/auth0/java-jwt/issues/511#issuecomment-914228182)
- In the issue, the user reported that session data works as expected, but I'll extend our tests to cover this.

src/test/java/io/supertokens/test/session/api/RefreshSessionAPITest2_7.java
porcellus: `checkRefreshSessionResponse` called on 393 already checks this.

rishabhpoddar (approved):

PR #503, opened 2022-09-06
feat: adds migration api and queries

jscyo:
## Summary of change
(A few sentences about this PR)

## Related issues
- Link to issue1 here
- Link to issue1 here

## Test Plan
(Write your test plan here. If you changed any code, please provide us with clear instructions on how you verified your changes work. Bonus points for screenshots and videos!)

## Documentation changes
(If relevant, please create a PR in our [docs repo](https://github.com/supertokens/docs), or create a checklist here highlighting the necessary changes)

## Checklist for important updates
- [ ] Changelog has been updated
    - [ ] If there are any db schema changes, mention those changes clearly
- [ ] `coreDriverInterfaceSupported.json` file has been updated (if needed)
- [ ] `pluginInterfaceSupported.json` file has been updated (if needed)
- [ ] Changes to the version if needed
   - In `build.gradle`
- [ ] Had installed and ran the pre-commit hook
- [ ] If there are new dependencies that have been added in `build.gradle`, please make sure to add them in `implementationDependencies.json`.
- [ ] Issue this PR against the latest non released version branch.
   - To know which one it is, run find the latest released tag (`git tag`) in the format `vX.Y.Z`, and then find the latest branch (`git branch --all`) whose `X.Y` is greater than the latest released tag.
   - If no such branch exists, then create one from the latest released branch.

## Remaining TODOs for this PR
- [ ] Convert all flows to tests
- [ ] Add firebase Scrypt support
   - [ ] How do other providers to password hash migration from firebase
   - [ ] what prefix do we add to the password hash
   - [ ] Does firebase always give the same length password hash or is it configurable
   - [ ] add tests for verification 
- [ ] see if firebase allows you to change the key, rounds.. config. If they do, how to manage backward compatibility? If they don't allow you to change, have they changed it in the past and have they maintained backwards compatability
- [ ] What is the nature of firebase verification, is it similar to argon? Do we need to use pool size like with argon? If it is do we need to give a config to change the pool size.
- [ ] speak to Mihaly about Scrypt and encryption(Cost and edge case etc...).

PR #507, opened 2022-09-13
feat: Adds support to migrate users from Firebase

jscyo:
## Summary of change
Adds support for verifying Firebase SCrypt password hashes

## Related issues
- https://github.com/supertokens/supertokens-core/pull/503

## Test Plan
(Write your test plan here. If you changed any code, please provide us with clear instructions on how you verified your changes work. Bonus points for screenshots and videos!)

## Documentation changes
(If relevant, please create a PR in our [docs repo](https://github.com/supertokens/docs), or create a checklist here highlighting the necessary changes)

   - In `build.gradle`

- [x] update emailpassword_users table, change passwordHash column size, also do for other plugins
- [x] see if firebase allows you to change the key, rounds.. config. If they do, how to manage backward compatibility? If they don't allow you to change, have they changed it in the past, and have they maintained backward compatibility.
   - Firebase does not allow you to change the hash config. From research, it seems they have not updated the config values from as early as 2015.
- [x] What is the nature of firebase verification, is it similar to argon? Do we need to use pool size like with argon? If it is do we need to give a config to change the pool size.
- [x] speak to Mihaly about Scrypt and encryption(Cost and edge case etc...).


jscyo:
### Format for Firebase SCrypt hash we are storing:
f_scrypt$XyOcS6THky2UOYTS5pQy7e8Ig1V0Syabg16FpoSgmV4erfBg7qkwlOvcvf9l8GldcpIus1oV3XNdIPCMw06+zg==$2PcV2yt+oXPwag==$m14$r11$sBw==

### Structure
Prefix: `f_scrypt`
split hash with `$`
splitHash.length == 6
splitHash[1] : password hash
splitHash[2]: salt
splitHash[3]: memory cost; hash prefix `m`
splitHash[4]: number of rounds; has prefix `r`
splitHash[5]: salt separator; has prefix `s`

build.gradle
```diff
@@ -61,6 +61,9 @@ dependencies {
     // https://mvnrepository.com/artifact/com.auth0/java-jwt
     implementation 'com.auth0:java-jwt:4.0.0'
 
+    // https://mvnrepository.com/artifact/com.lambdaworks/scrypt
+    implementation group: 'com.lambdaworks', name: 'scrypt', version: '1.4.0'
```
rishabhpoddar, line 65: you need to also modify implementation-dependencies.json file..

config.yaml
```diff
@@ -105,4 +105,10 @@ core_config_version: 0
 # argon2_hashing_pool_size:
 
 # (OPTIONAL | Default: "INFO"). Logging level for the core. Values are "DEBUG" | "INFO" | "WARN" | "ERROR" | "NONE"
-# log_level:
\ No newline at end of file
+# log_level:
+
+# (OPTIONAL | Default: null). The signing key used for firebase scrypt password hashing
+# firebase_signing_key:
```
rishabhpoddar, line 111:
```suggestion
# firebase_password_hashing_signing_key:
```

rishabhpoddar: Also, is it a signing key for sure? Where does it say that in the firebase docs?

config.yaml
```diff
@@ -105,4 +105,10 @@ core_config_version: 0
...
+
+# (OPTIONAL | Default: 1). Number of concurrent firebase scrypt hashes that can happen at the same time for sign in requests.
+# firebase_hashing_pool_size:
```
rishabhpoddar, line 114:
```suggestion
# firebase_password_hashing_pool_size:
```

src/main/java/io/supertokens/emailpassword/PasswordHashing.java
```diff
@@ -113,32 +132,32 @@ public boolean verifyPasswordWithHash(String password, String hash) {
...
             }
         } else if (PasswordHashingUtils.isInputHashInBcryptFormat(hash)) {
             ProcessState.getInstance(main).addState(ProcessState.PROCESS_STATE.PASSWORD_VERIFY_BCRYPT, null);
             String bCryptPasswordHash = PasswordHashingUtils
                     .replaceUnsupportedIdentifierForBcryptPasswordHashVerification(hash);
             return BCrypt.checkpw(password, bCryptPasswordHash);
+        } else if (PasswordHashingUtils.isInputHashInFirebaseSCryptFormat(hash)) {
+            ProcessState.getInstance(main).addState(ProcessState.PROCESS_STATE.PASSWORD_VERIFY_FIREBASE_SCRYPT, null);
+            return withFirebaseSCryptConcurrencyLimited(() -> PasswordHashingUtils
+                    .verifyFirebaseSCryptPasswordHash(password, hash, Config.getConfig(main).getFirebaseSigningKey()));
```
rishabhpoddar, line 153: you do not have a null check for the value returned by getFirebaseSigningKey. Nor is there a null check for that in the config class

src/main/java/io/supertokens/config/CoreConfig.java
```diff
@@ -188,6 +198,10 @@ public int getArgon2Parallelism() {
         return argon2_parallelism;
     }
 
+    public String getFirebaseSigningKey() {
+        return firebase_signing_key;
```
rishabhpoddar, line 202: what if this is null when the function is called?

src/main/java/io/supertokens/emailpassword/PasswordHashing.java
```diff
@@ -89,22 +91,39 @@ public String createHashWithSalt(String password) {
...
+                try {
+                    // put will wait for there to be an empty slot in the queue and return
+                    // only when there is a slot for waiter
+                    this.firebaseSCryptBoundedQueue.put(waiter);
+                } catch (InterruptedException ignored) {
+                }
+            }
+            return func.op();
+        } finally {
+            this.argon2BoundedQueue.remove(waiter);
```
rishabhpoddar, line 126: - why argon2 here??
- Why have you made two functions for the same thing?? Just pass in the queue as an arg to one common function.

src/main/java/io/supertokens/emailpassword/PasswordHashing.java
```diff
@@ -113,32 +132,32 @@ public boolean verifyPasswordWithHash(String password, String hash) {
...
-//        }
 
         return false;
     }
 
     @TestOnly
     public int getBlockedQueueSize() {
-        return this.boundedQueue.size();
+        return this.argon2BoundedQueue.size();
     }
```
rishabhpoddar, line 162: change name to getArgon2BlockedQueueSize

src/main/java/io/supertokens/emailpassword/PasswordHashingUtils.java
```diff
@@ -17,12 +17,43 @@
...
+            this.passwordHash = passwordHash;
+            this.salt = salt;
+            this.saltSeparator = saltSeparator;
+            this.rounds = rounds;
+            this.memCost = memCost;
+        }
+    }
+
+    private static final String FIREBASE_SCRYPT_PREFIX = "f_scrypt";
+    private static final String FIREBASE_SCRYPT_SEPARATOR = "\\$";
```
rishabhpoddar, line 52: why the `\\`?

src/main/java/io/supertokens/emailpassword/PasswordHashingUtils.java
```diff
@@ -57,11 +88,12 @@ public static void assertSuperTokensSupportInputPasswordHashFormat(String passwo
             }
         }
 
-//        if (hashingAlgorithm.equals(PasswordHashingAlgorithm.SCRYPT)) {
-//            if (!isInputHashInScryptFormat(passwordHash)) {
-//                throw new UnsupportedPasswordHashingFormatException("Password hash is in invalid SCrypt format");
-//            }
-//        }
+        if (hashingAlgorithm.equals(CoreConfig.PASSWORD_HASHING_ALG.FIREBASE_SCRYPT)) {
```
rishabhpoddar, line 91: you missed out on changing the if condition right on top of this function.

src/main/java/io/supertokens/emailpassword/PasswordHashingUtils.java
```diff
@@ -70,26 +102,15 @@ public static String updatePasswordHashWithPrefixIfRequired(String passwordHash,
...
         }
 
-//        if (hashingAlgorithm == PasswordHashingAlgorithm.SCRYPT){
-//            if (doesPasswordHashHaveScrpytPrefix(passwordHash)){
-//                return passwordHash;
-//            }
-//            return addScryptPrefixToPasswordHash(passwordHash);
-//        }
-
+        if (hashingAlgorithm == CoreConfig.PASSWORD_HASHING_ALG.FIREBASE_SCRYPT) {
```
rishabhpoddar, line 105: im questioning if it still makes sense to have this function - since we will be telling users to give the hashes in a specific format anyway, so we can ask them to add the desired prefix as well

src/main/java/io/supertokens/emailpassword/PasswordHashingUtils.java
```diff
@@ -101,13 +122,139 @@ public static boolean isInputHashInArgon2Format(String hash) {
...
+            boolean containsSaltSeparator = false;
+
+            for (int i = 3; i < separatedHash.length; i++) {
+                // does hash contain memory cost
+                if (separatedHash[i].startsWith(FIREBASE_SCRYPT_MEM_COST_SEPARATOR)) {
+                    // check that memory cost is a number
+                    try {
+                        Integer.parseInt(separatedHash[3].split(FIREBASE_SCRYPT_MEM_COST_SEPARATOR)[1]);
+                        containsMemCost = true;
+                    } catch (NumberFormatException e) {
```
rishabhpoddar, line 147: You should just catch all exceptions here.. what if `[1]` index doesn';t exist?

src/main/java/io/supertokens/emailpassword/PasswordHashingUtils.java
```diff
@@ -101,13 +122,139 @@ public static boolean isInputHashInArgon2Format(String hash) {
...
+                    }
+                }
+
+                // does hash contain rounds
+                if (separatedHash[i].startsWith(FIREBASE_SCRYPT_ROUNDS_SEPARATOR)) {
+                    // check that rounds is a number
+                    try {
+                        Integer.parseInt(separatedHash[4].split(FIREBASE_SCRYPT_ROUNDS_SEPARATOR)[1]);
+                        containsRounds = true;
+                    } catch (NumberFormatException e) {
```
rishabhpoddar, line 158: catch all exceptions

src/main/java/io/supertokens/emailpassword/PasswordHashingUtils.java
```diff
@@ -101,13 +122,139 @@ public static boolean isInputHashInArgon2Format(String hash) {
...
+                        return false;
```
rishabhpoddar, line 159: no need to return false here cause you won't be setting `containsRounds = true;` anyway, so it will fail the condition after the for loop.

src/main/java/io/supertokens/emailpassword/PasswordHashingUtils.java
```diff
@@ -101,13 +122,139 @@ public static boolean isInputHashInArgon2Format(String hash) {
...
+
+            for (int i = 3; i < separatedHash.length; i++) {
+                // does hash contain memory cost
+                if (separatedHash[i].startsWith(FIREBASE_SCRYPT_MEM_COST_SEPARATOR)) {
+                    // check that memory cost is a number
+                    try {
+                        Integer.parseInt(separatedHash[3].split(FIREBASE_SCRYPT_MEM_COST_SEPARATOR)[1]);
+                        containsMemCost = true;
+                    } catch (NumberFormatException e) {
+                        return false;
```
rishabhpoddar, line 148: no need to return false here cause you won't be setting `containsMemCost = true;` anyway, so it will fail the condition after the for loop.

src/main/java/io/supertokens/emailpassword/PasswordHashingUtils.java
```diff
@@ -17,12 +17,43 @@
...
+                int memCost) {
+            this.passwordHash = passwordHash;
+            this.salt = salt;
+            this.saltSeparator = saltSeparator;
+            this.rounds = rounds;
+            this.memCost = memCost;
+        }
+    }
+
+    private static final String FIREBASE_SCRYPT_PREFIX = "f_scrypt";
```
rishabhpoddar, line 51: should be prefix also start with a `$`? What do bcrypt and argon2 do here?

src/main/java/io/supertokens/emailpassword/PasswordHashingUtils.java
```diff
@@ -101,13 +122,139 @@ public static boolean isInputHashInArgon2Format(String hash) {
...
+    private static boolean doesPasswordHashHaveFireBaseSCryptPrefix(String passwordHash) {
+        return passwordHash.startsWith(FIREBASE_SCRYPT_PREFIX);
+    }
+
+    private static String addFirebaseSCryptPrefixToPasswordHash(String passwordHash) {
+        return FIREBASE_SCRYPT_PREFIX + "$" + passwordHash;
+    }
+
+    private static ParsedFireBaseSCryptResponse retrieveFirebaseSCryptInfoFromStoredPasswordHash(String hash) {
+
```
rishabhpoddar, line 182: this function and isInputHashInFirebaseSCryptFormat can be combined into one.. need refactoring.

src/main/java/io/supertokens/emailpassword/PasswordHashingUtils.java
```diff
@@ -101,13 +122,139 @@ public static boolean isInputHashInArgon2Format(String hash) {
...
-//        // TODO:
-//        return passwordHash;
-//    }
+        ParsedFireBaseSCryptResponse response = retrieveFirebaseSCryptInfoFromStoredPasswordHash(passwordHash);
+
+        passwordHash = response.passwordHash;
+        String salt = response.salt;
+        int memCost = response.memCost;
+        int rounds = response.rounds;
+        String saltSep = response.saltSeparator;
```
rishabhpoddar, line 220: why this.. just use response. below..

src/main/java/io/supertokens/emailpassword/PasswordHashingUtils.java
```diff
@@ -101,13 +122,139 @@ public static boolean isInputHashInArgon2Format(String hash) {
...
+            hashedBytes = SCrypt.scrypt(plainTextPassword.getBytes(StandardCharsets.US_ASCII), saltConcat, N, rounds, p,
+                    64);
+        } catch (GeneralSecurityException e) {
+            return false;
+        }
+        // encrypting with aes
+        byte[] signerBytes = Base64.decodeBase64(base64_signer_key.getBytes(StandardCharsets.US_ASCII));
+        byte[] cipherTextBytes = encrypt(signerBytes, hashedBytes);
+
+        return new String(Objects.requireNonNull(Base64.encodeBase64(cipherTextBytes))).equals(passwordHash);
```
rishabhpoddar, line 245: where did you get all this code from?

src/main/java/io/supertokens/emailpassword/PasswordHashingUtils.java
```diff
@@ -101,13 +122,139 @@ public static boolean isInputHashInArgon2Format(String hash) {
...
+    }
+
+    private static byte[] encrypt(byte[] signer, byte[] derivedKey) {
```
rishabhpoddar, line 248: Why does this need to be in a separate function? It's called only in one place

src/test/java/io/supertokens/test/emailpassword/PasswordHashingTest.java
```diff
@@ -16,11 +16,14 @@
 
 package io.supertokens.test.emailpassword;
 
+import com.lambdaworks.crypto.SCrypt;
```
rishabhpoddar, line 19: Test cases you need to cover:
- [ ] Argon2 / bcrypt hashing works with all the various types
- [ ] Firebase input hash not meeting right format in API yields a 400 error response (try all combinations of incorrect input)
- [ ] Once firebase password hash is saved, verifying the passwords works.
- [ ] If we are keeping the modifying prefix thing, the flow should work even if the firebase hash doesn't have the prefix.
- [ ] Doing several firebase hashes (similar to how we did for argon2), shouldn't yield the server to crash - this is testing the pool size thing
- [ ] Not providing the firebase signing key, but providing a firebase hash in the import API returns a 400 error.
- [ ] Not providing the firebase signing key, but trying to sign in with a firebase password should yield a 400 error.
- [ ] Providing a wrong firebase signing key and trying to sign in with a firebase password should yield invalid credentials status response. Here the signing key should be a a random string.
- [ ] Store two different firebase hashes with different memcost / rounds etc and make sure that verifying the password against those hashes still works.
- [ ] Providing wrong password during sign in with firebase password should yield invalid credentials status response.
- [ ] If we have stored the firebase password hash is incorrectly stored in the db, and then trying to sign in with the right password should give invalid credentials status response.
- [ ] Jumbling the oder of the rounds, memory, salt separator should still make things work.
- [ ] What happens is salt separator contains a `$` or `$$`, or `$dsf$$` or some other combination of using `$`? Can we still make it work?

src/main/java/io/supertokens/emailpassword/PasswordHashingUtils.java
```diff
@@ -17,12 +17,43 @@
...
+import javax.crypto.Cipher;
+import javax.crypto.spec.IvParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
+import java.nio.charset.StandardCharsets;
+import java.security.*;
+import java.util.Objects;
 
 public class PasswordHashingUtils {
 
+    public static class ParsedFireBaseSCryptResponse {
```
rishabhpoddar, line 34: add a static function `fromHashString(String hash)` and which returns `null` or throws an error in case you can't parse it, else return an instance of `ParsedFireBaseSCryptResponse`.

Make this class its own file.

Add unit tests for this class separately, independent of anything else.

rishabhpoddar (changes requested):

config.yaml
jscyo: In firebase's docs, the key is `base64_signer_key`, with the description as "The signer key from the project's password hash parameters"

src/main/java/io/supertokens/emailpassword/PasswordHashingUtils.java
jscyo: `split` takes a regex, so `$` needs to be escaped

src/main/java/io/supertokens/emailpassword/PasswordHashingUtils.java
jscyo: For bcrypt, they start with the prefix`$2` followed by the version which are `$2a`, `$2b`...
For argon2, they start with prefix `$argon2id`, `$argon2i`, `$argon2d`

We could add it sure

jscyo: updated the prefix to now be `$f_scrypt`

src/main/java/io/supertokens/config/CoreConfig.java
```diff
@@ -188,6 +198,13 @@ public int getArgon2Parallelism() {
         return argon2_parallelism;
     }
 
+    public String getFirebase_password_hashing_signer_key() {
+        if (firebase_password_hashing_signer_key == null) {
+            throw new QuitProgramException("'firebase_password_hashing_signer_key' cannot be null");
```
rishabhpoddar, line 203: this is bad..

rishabhpoddar: Just throw a new IllegalStateException -> which should yield a 500 error in sign in request (add a test for this as well)

src/main/java/io/supertokens/emailpassword/PasswordHashingUtils.java
```diff
@@ -57,11 +67,12 @@ public static void assertSuperTokensSupportInputPasswordHashFormat(String passwo
...
-//        }
+        if (hashingAlgorithm.equals(CoreConfig.PASSWORD_HASHING_ALG.FIREBASE_SCRYPT)) {
+            if (ParsedFirebaseSCryptResponse.fromHashString(passwordHash) == null) {
+                throw new UnsupportedPasswordHashingFormatException(
+                        "Password hash is in invalid Firebase SCrypt format");
+            }
+        }
     }
 
     public static String updatePasswordHashWithPrefixIfRequired(String passwordHash,
```
rishabhpoddar, line 78: remove the use of this function

src/main/java/io/supertokens/emailpassword/PasswordHashingUtils.java
```diff
@@ -101,13 +101,55 @@ public static boolean isInputHashInArgon2Format(String hash) {
         return (hash.startsWith("$argon2id") || hash.startsWith("$argon2i") || hash.startsWith("$argon2d"));
     }
 
-//    private static  boolean doesPasswordHashHaveScrpytPrefix(String passwordHash){
-//        // TODO:
-//        return false;
-//    }
+    private static boolean doesPasswordHashHaveFireBaseSCryptPrefix(String passwordHash) {
+        return passwordHash.startsWith(FIREBASE_SCRYPT_PREFIX);
+    }
```
rishabhpoddar, line 106: remove this as well

src/main/java/io/supertokens/emailpassword/PasswordHashingUtils.java
```diff
@@ -101,13 +101,55 @@ public static boolean isInputHashInArgon2Format(String hash) {
...
+
+    private static String addFirebaseSCryptPrefixToPasswordHash(String passwordHash) {
+        return FIREBASE_SCRYPT_PREFIX + "$" + passwordHash;
+    }
```
rishabhpoddar, line 110: not requirerd

src/main/java/io/supertokens/emailpassword/PasswordHashingUtils.java
```diff
@@ -101,13 +101,55 @@ public static boolean isInputHashInArgon2Format(String hash) {
...
+
+    public static boolean verifyFirebaseSCryptPasswordHash(String plainTextPassword, String passwordHash,
+            String base64_signer_key) {
+
```
rishabhpoddar, line 114: comment a link to the source of the below code

src/main/java/io/supertokens/emailpassword/PasswordHashingUtils.java
```diff
@@ -17,12 +17,22 @@
...
+import javax.crypto.Cipher;
+import javax.crypto.spec.IvParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
+import java.nio.charset.StandardCharsets;
+import java.security.*;
+import java.util.Objects;
 
 public class PasswordHashingUtils {
 
+    private static final String FIREBASE_SCRYPT_PREFIX = "$f_scrypt";
```
rishabhpoddar, line 34: this should go into ParsedFirebaseSCryptResponse.java file

src/main/java/io/supertokens/emailpassword/ParsedFirebaseSCryptResponse.java
```diff
@@ -0,0 +1,74 @@
...
+            String[] separatedPasswordHash = hash.split(FIREBASE_SCRYPT_SEPARATOR);
+            if (separatedPasswordHash.length != 7) {
+                return null;
+            }
+            String passwordHash = separatedPasswordHash[2];
+            String salt = separatedPasswordHash[3];
+            String saltSeparator = null;
+            Integer memCost = null;
+            Integer rounds = null;
+
```
rishabhpoddar, line 51: you need to check that the first element of the split array is `FIREBASE_SCRYPT_PREFIX`

src/main/java/io/supertokens/emailpassword/ParsedFirebaseSCryptResponse.java
```diff
@@ -0,0 +1,74 @@
...
+                if (separatedPasswordHash[i].startsWith(FIREBASE_SCRYPT_SALT_SEPARATOR)) {
+                    saltSeparator = separatedPasswordHash[i].split(FIREBASE_SCRYPT_SALT_SEPARATOR)[1];
+                }
+            }
+
+            if (passwordHash == null || salt == null || saltSeparator == null || memCost == null || rounds == null) {
+                return null;
+            }
+            return new ParsedFirebaseSCryptResponse(passwordHash, salt, saltSeparator, rounds, memCost);
+        } catch (Exception e) {
```
rishabhpoddar, line 70: catch Throwable instead of exception

src/main/java/io/supertokens/emailpassword/ParsedFirebaseSCryptResponse.java
```diff
@@ -40,9 +41,12 @@ public ParsedFirebaseSCryptResponse(String passwordHash, String salt, String sal
     public static ParsedFirebaseSCryptResponse fromHashString(String hash) {
         try {
             String[] separatedPasswordHash = hash.split(FIREBASE_SCRYPT_SEPARATOR);
-            if (separatedPasswordHash.length != 7) {
+
+            // check that stored password hash continas 7 fields and the first field has the firebase scrypt prefix
+            if (!(separatedPasswordHash.length == 7 && separatedPasswordHash[1].equals(FIREBASE_SCRYPT_PREFIX))) {
```
rishabhpoddar, line 46: check that index 0 is `""`

rishabhpoddar (changes requested):

PR #514, opened 2022-09-30
fix: correct db migration statement for postgres

n1ru4l:
The MySQL statement was duplicated. For Postgres `ALTER COLUMN` must be used instead of `MODIFY`


rishabhpoddar (approved):

rishabhpoddar:
Thanks for the fix @n1ru4l

PR #576, opened 2023-03-01
refactor: adding/using a semver class instead of version strings

porcellus:
## Summary of change

Adding/using a SemVer class instead of version strings. This will used in #566 to enable version ranges instead of equality checks for CDI versions

## Related issues

- Link to issue1 here
- Link to issue1 here

## Test Plan

Existing tests should cover everything

## Documentation changes

N/A

    - In `build.gradle`
- [x] If there are new dependencies that have been added in `build.gradle`, please make sure to add them

### DIFF
--- a/src/main/java/io/supertokens/utils/SemVer.java
+++ b/src/main/java/io/supertokens/utils/SemVer.java
@@ -39,40 +39,67 @@ public class SemVer implements Comparable<SemVer> {
     }
 
     public SemVer(String version) {
-        if(version == null)
+        if(version == null) {
             throw new IllegalArgumentException("Version can not be null");
-        if(!version.matches("[0-9]+(\\.[0-9]+)*"))
+        }
+
+        if(!version.matches("[0-9]+(\\.[0-9]+)*")) {
             throw new IllegalArgumentException("Invalid version format");
+        }
+
         this.version = version;
     }
 
+    public boolean betweenInclusive(SemVer min, SemVer max) {
+        return min.compareTo(this) <= 0 && this.compareTo(max) <= 0;
+    }
+
+    public boolean greaterThanOrEqualTo(SemVer min) {
+        return min.compareTo(this) <= 0;
+    }
+
+    public boolean lesserThan(SemVer max) {
+        return this.compareTo(max) < 0;
+    }
+
     @Override public int compareTo(SemVer that) {
-        if(that == null)
+        if (that == null) {
             return 1;
+        }
+
         String[] thisParts = this.get().split("\\.");
         String[] thatParts = that.get().split("\\.");
+
         int length = Math.max(thisParts.length, thatParts.length);
-        for(int i = 0; i < length; i++) {
-            int thisPart = i < thisParts.length ?
-                    Integer.parseInt(thisParts[i]) : 0;
-            int thatPart = i < thatParts.length ?
-                    Integer.parseInt(thatParts[i]) : 0;
-            if(thisPart < thatPart)
+
+        for (int i = 0; i < length; i++) {
+            int thisPart = i < thisParts.length ? Integer.parseInt(thisParts[i]) : 0;
+            int thatPart = i < thatParts.length ? Integer.parseInt(thatParts[i]) : 0;
+
+            if(thisPart < thatPart) {
                 return -1;
-            if(thisPart > thatPart)
+            }
+
+            if(thisPart > thatPart) {
                 return 1;
+            }
         }
+
         return 0;
     }
 
     @Override public boolean equals(Object that) {
-        if(this == that)
+        if(this == that) {
             return true;
-        if(that == null)
-            return false;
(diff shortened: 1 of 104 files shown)


src/main/java/io/supertokens/utils/SemVer.java
```diff
@@ -0,0 +1,88 @@
...
+            return false;
+
+        if(this.getClass() != that.getClass())
+            return false;
+        return this.compareTo((SemVer) that) == 0;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(version);
```
rishabhpoddar, line 81:
```suggestion
        return version.hashCode();
```

src/main/java/io/supertokens/utils/SemVer.java
```diff
@@ -0,0 +1,88 @@
...
+        return 0;
+    }
+
+    @Override public boolean equals(Object that) {
+        if(this == that)
+            return true;
+        if(that == null)
+            return false;
+
+        if(this.getClass() != that.getClass())
```
rishabhpoddar, line 74: Use instanceof instead

src/main/java/io/supertokens/webserver/api/emailpassword/SignUpAPI.java
```diff
@@ -70,9 +70,6 @@ protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws I
             JsonObject result = new JsonObject();
             result.addProperty("status", "OK");
             JsonObject userJson = new JsonParser().parse(new Gson().toJson(user)).getAsJsonObject();
-            if (super.getVersionFromRequest(req).equals("2.4")) {
-                userJson.remove("timeJoined");
-            }
```
rishabhpoddar, line 75: why is this removed?

src/main/java/io/supertokens/webserver/api/emailpassword/UserAPI.java
```diff
@@ -103,9 +103,6 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws IO
                 JsonObject result = new JsonObject();
                 result.addProperty("status", "OK");
                 JsonObject userJson = new JsonParser().parse(new Gson().toJson(user)).getAsJsonObject();
-                if (super.getVersionFromRequest(req).equals("2.4")) {
-                    userJson.remove("timeJoined");
-                }
```
rishabhpoddar, line 108: why is this removed?

src/main/java/io/supertokens/utils/SemVer.java
```diff
@@ -0,0 +1,88 @@
...
+
+    public SemVer(String version) {
+        if(version == null)
+            throw new IllegalArgumentException("Version can not be null");
+        if(!version.matches("[0-9]+(\\.[0-9]+)*"))
+            throw new IllegalArgumentException("Invalid version format");
+        this.version = version;
+    }
+
+    @Override public int compareTo(SemVer that) {
```
rishabhpoddar, line 49: - You should add extensive unit tests for this.
- Also, this function can be very easily misused.. it's not clear what it means if this function returns `-1`. Instead, i would prefer to have a function like isGreaterThan, isLesserThan, isEqualTo

rishabhpoddar (changes requested):

src/main/java/io/supertokens/webserver/api/emailpassword/SignUpAPI.java
porcellus: We no longer support `2.4`, (2.7 is the lowest) right now.

src/main/java/io/supertokens/webserver/api/emailpassword/UserAPI.java
porcellus: We no longer support `2.4`, (2.7 is the lowest) right now.

src/main/java/io/supertokens/utils/SemVer.java
porcellus: This is the default `Comparable` interface from java that I think we should definitely implement. I'll add those functions in the other core PR since they will be used there. I debated adding them here and replacing equality checks, but I wanted to make this PR with minimal logic changes.
Btw, we basically only need:
- `equals`
- `atLeast` (or `greaterThanOrEqualTo`)
- `betweenInclusive` (for version ranges like `2.9`-`2.18`)
with maybe a `lesserThan` (but not equal) in the future

rishabhpoddar: Yea. Please implement greaterThanOrEqualTo, betweenInclusive and lesserThan

rishabhpoddar (changes requested):

rishabhpoddar (approved):

rishabhpoddar (approved):

Issue #554, opened 2023-01-12
TOTP / MFA login discussions


rishabhpoddar:
Google doc link: https://docs.google.com/document/d/1qyd4XmepLJXNC4uURcak-7JbWP0UuzOPSsjBInVAuNk/edit

## TODO:
- [x] Consider seperately mentioning templates vars used in qr (issuerName, secret, provisioner, etc)
- [ ] ~~Fetch user identifier based on first factor?~~
- [ ] ~~Give Rishabh all the touch points between account linking and mfa recipe~~
- [x] ~~Think if there are any session changes required for TOTP verify code and verify device API? ~~
- [x] How to prevent the attack in which the attacker finishes the fist factor and then for second factor (which is totp), they call the create new device API and finish the flow with their new device - even though the user already has an existing device. One idea is to add a claim validator to the create device API that all MFA claims should be completed when this API is called. What other solution is there? Some edge cases with this solution:
   - How will this solution affect the sign up process?
   - What if the user signs up with first factor, logs out, and then signs in  - that time we still want to allow them to set the factor even though it's during sign in.
   - [ ] ~~We decided to prevent this by checking if all factors are completed in the `createDevicePOST` implementation (claim assertion), if there already exists one device. But we should discuss this with the team (TODO)~~
- [ ] ~~We need to get rid of TOTP_NOT_ENABLED_ERROR from the core entirely:~~
   - verifyCode should return  INVALID_TOTP_ERROR
   - verifyDevice should return UNKNOWN_DEVICE_ERROR
   - updateDevice should return UNKNOWN_DEVICE_ERROR
   - removeDevice should return { status: "OK"; didDeviceExist: false }
   - listDevices should return an empty array
- [ ] ~~Need to have a separate API for MFA account linking which is enabled if the MFA feature is enabled and account linking is not. This API will only allow account linking for the purpose of MFA. So you need to enforce that, somehow.~~
   - [ ] ~~One idea is that we disallow linking of accounts across the same recipe ID.~~
- [ ] ~~Devs should be able to set MFA settings for tenants via an API call (similar to how they configure third party providers for tenants via an API call). This should be as flexible as possible.~~
   - [ ] ~~Maybe we can have enableFactorForTenant(tenantId, factorId…) just like we have enableFactorForUser. This is not necessary.. it's just an idea~~
- [x] Figure out changes in core, backend sdk and frontend sdk related to this: https://github.com/supertokens/supertokens-core/issues/554#issuecomment-1748645306
[...]

PR #586, opened 2023-03-21
test: Add tests for active users update across different API calls

KShivendu:
## Summary of change

Add tests for active users update across different API calls

## Related issues

- https://github.com/supertokens/supertokens-core/issues/554

    - In `build.gradle`

### DIFF
--- a/src/main/java/io/supertokens/ActiveUsers.java
+++ b/src/main/java/io/supertokens/ActiveUsers.java
@@ -11,4 +11,8 @@ public class ActiveUsers {
         } catch (StorageQueryException ignored) {
         }
     }
+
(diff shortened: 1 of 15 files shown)


src/main/java/io/supertokens/ActiveUsers.java
```diff
@@ -11,4 +11,8 @@ public static void updateLastActive(Main main, String userId) {
         } catch (StorageQueryException ignored) {
         }
```
rishabhpoddar, line 12: - in all the relevant APIs, first send the response, and then call the last update time (first to a test for 20 seconds waiting)
- in session apis, the part where you are calling a user id mapping, it needs to be in a try / catch.

rishabhpoddar (changes requested):

Issue #447, opened 2022-05-15
Implement account linking


rishabhpoddar:
This issue is about account linking

Flow diagrams: https://lucid.app/lucidchart/82064b11-858b-4f97-b2ee-3d6e6ed604a6/edit?viewport_loc=-91%2C-655%2C2048%2C1196%2C0_0&invitationId=inv_92259a69-24b4-472f-bbc6-0af69bf835a5#

## TODO:
- [x] All recipes in backend sdk
   - [x] emailpassword
   - [x] thirdpartyemailpassword
   - [x] thirdparty
   - [x] passwordless
   - [x] thirdpartypasswordless
   - [x] session
   - [x] metadata
   - [x] dashboard
   - [x] jwt
   - [x] openid
   - [x] emailverification
   - [x] roles
   - [x] useridmapping
- [x] Fix typing for `linkAccountToExistingAccountPOST` API on backend SDK
- [x] In `linkAccountsWithUserFromSession` should we throw email verification claim failure in case the existing account is not verified? How will this be handled on the frontend (with respect to claim validation order for mfa: chat reference: https://supertokens.slack.com/archives/D02B887A2MQ/p1683003076081709)? 
- [x] email verification functions should be given recipe user ID only (internal to our SDK)
- [x] Make email verification functions take email compulsory - so that if a primary user ID is passed, it does not end up getting the wrong email (we should also make an ADR for this. See comment [here](https://github.com/supertokens/supertokens-node/pull/439#discussion_r1032377777))
  - We've decided against doing this, the EV functions take a strictly typed recipeUserId instead.
- [x] Make a demo app to show user how to add phonerNumber or email to an existing logged in account. E.g. user logged in via TP flow, and on settings page they add password or phoneNumber and it should create emailpassword or passwordless account 
- [x] Fix dashboard getUsers function to return the new user type. This also involves changing the frontend
- [x] Function name changes:
    - [x] getUserByAccountInfo (current) to getUserByAccountInfoWithLoginMethod (new)
- [ ] Review core paths and methods for the new APIs
- [ ] Fix frontend of dashboard to work with new user object type
- [x] Try and minimise use of user type specific to each recipe, and use the main User object as much as possible. For example, in sign up function / API for emailpassword recipe, we want to return the main User object as opposed to the emailpassword user object - right?
- [x] How would someone customise it to have account linking with a prompt during sign in?
  - This will be included as an example. In short: they can add a claim somewhat similar to the proposed account linking claim.
[...]

PR #742, opened 2023-07-07
feat: account linking

rishabhpoddar:
## Summary of change

Base PR for account linking related changes in the core

## Related issues

- https://github.com/supertokens/supertokens-core/issues/447

## Test Plan

## Documentation changes

    - In `build.gradle`

- [x] Check query performance for pagination and user count cause we use group by in it now.
- [x] Check if both indices - `all_auth_recipe_users_primary_user_id_index` and `all_auth_recipe_users_primary_user_id_and_tenant_id_index` are needed
- [x] Tests for user exists where we return true if primary user exists, but it's associated recipe user doesn't and the primary user is linked to other recipe user.
- [x] User pagination changes.
- [x] Remove UserInfo object from plugin interface that inherit from AuthRecipeUserInfo
- [x] Add an equals function in LoginMethods and AuthRecipeUserInfo in plugin interface
- [x] remove `getRecipeId` from AuthRecipeUserInfo class
- [x] Check if separate index is needed for (app_id, user_id) in `all_auth_recipe_users` given that there already exists a fk on it, and we use this in the getUsers function.
- [x] Remove input of `recipeId` from `getUserInfoForRecipeIdFromUserIds`
- [x] We should not longer use recipe level functions like getThirdPartyUserInfoUsingId.
- [x] Implement hashCode function for AuthRecipeUserInfo and LoginMethods
- [x] When creating user object in queries implementations, we pass `false` to isPrimary everywhere. Need to actually fetch the value from the db. You can go to UserInfo obj in all the classes and then see where the constructor is being used.
- [ ] Optimise the fetching of userInfo object - instead of doing three queries, to fetch isPrimary, email verification and login info, try doing it in one.
- [x] Make tenantIds in loginMethods and AuthRecipeUserInfo into a set?
- [x] Make CDI 4.0 latest in SemVer
- [x] Implement toJson in AuthRecipeUserInfo
- [ ] Check how can we reduce number of queries done by GeneralQueries.getPrimaryUserInfoForUserId
- [x] In listPrimaryUsersByEmail, in GeneralQueries.java, we need to implement all auth recipes
- [ ] In listPrimaryUsersByEmail, in GeneralQueries.java, we need a more efficient way to get users based on emails directly - too many queries.
- [x] check that we no longer need to use `fillUserInfoWithIsPrimaryUserBoolean_transaction` in the queries files anymore.
- [x] Optimise query for listing users by email / phone number to fetch primary_or_recipe_id instead of just user_id in the recipe level queries.
- [x] Active users should take into account linked users and only record the primary user
- [ ] Account linking APIs:
   - [x] getUser -> id -> user object
   - [x] listUsersByAccountInfo
   - [x] deleteUser
   - [x] unlinkAccount
   - [x] linkAccounts
   - [x] canLinkAccounts
   - [x] createPrimaryUser
   - [x] canCreatePrimaryUser
   - [ ] getUsers
- [ ] See query plan for queries in getPrimaryUserIdUsingEmail for tp, ep and pless recipe
- [x] When linking 2 accounts, (A, B), we need to check if they can be linked across ALL tenants in tenantsOf(A) union tenantsOf(B) belong to.
- [x] When sharing users across tenants, we need to make sure that the state of account linking is not ruined in that tenant.
- [ ] ~~Remove use of `getPrimaryUserInfoForUserIds` in generalqueries that does not take Connection con as an arg.~~
- [ ] Test that create new primary user with same email causes deadlock which is eventually auto resolved.
- [x] Do we need `all_auth_recipe_users_pagination_index` and `all_auth_recipe_users_primary_user_id_and_tenant_id_index` (the second one is the same, except that it's missing the time_joined_column)
- [x] The index column order in `all_auth_recipe_users_pagination_index` has the app_id and tenant_id at the end as opposed to at the start. Does this matter?
- [x] For user count function, try out window functions:
   ```
   SELECT max(unique_user_rows) from (SELECT ROW_NUMBER() OVER(PARTITION BY app_id, tenant_id, 
   primary_or_recipe_user_id ORDER BY primary_or_recipe_user_id) as unique_user_rows
   FROM all_auth_recipe_users where app_id = '...' AND tenant_id = '...')
   ```
- [ ] Instead of querying auth recipe tables one by one to fetch info from them (by email, for example), try and use the UNION query across them all to combine the results (in the select part of these, try and use the same number of colums across the tables):
   ```
   (select user_id, email from emailpassword_users where email = '...') union (select user_id, email from passwordless_users 
   where email = '...') union (select user_id, email from thirdparty_users where email = '...')
   ```
   But if there are different number of columns, see this: https://stackoverflow.com/questions/15867152/mysql-union-different-number-of-columns
- [ ] Instead of doing so many FOR UPDATES, maybe checkout advisory lock (https://www.postgresql.org/docs/current/explicit-locking.html#ADVISORY-LOCKS) for psql, and get_lock in mysql
- [x] Investigate slow down in test when running with psql - it takes 4-5 seconds per test at a minimum.
- [x] Delete branch in node repo -> account-linking-core-tests-tracking
- [x] Call to `revokeAllSessionsForUser` in `linkAccounts` and `unlinkAccounts` should pass in false to revoke sessions for only the recipe user id
- [ ] https://github.com/supertokens/supertokens-core/pull/742#issuecomment-1651893320
- [ ] https://github.com/supertokens/supertokens-core/pull/742#issuecomment-1651889327 
- [x] Modification to sharing users across tenants to take into account account linking rules
- [x] Make sure that no production code is calling testOnly code
- [x] check that older versions of the sign in etc APIs return the right user object
- [x] Need to add user id mapping for all loginMethods in the user object as well, and not just the primary user.
- [x] How does user id mapping behave when it's mapped to a primary user was deleted (but still has linked recipe users), and then you try and map another recipe user id to the same primary user ID, or the same external user ID.
- [x] make db migratino script specific to psql and mysql
- [x] confirm db migration script is correct for psql and mysql
- [x] Remove all uses of recipe.getUserId internally, for example, we still use Passwordless.getUserId in updateUser in passwordless recipe.
- [x] Check for use of deprecated method in non deprecated code, and check for use of testOnly function in production code. 
- [x] Add fkey constrain on primary_or_recipe_user_id to appid to user id table
- [x] Check that deleting a primary user ID (not unlinking it) while it's linked to other users, should NOT remove it from app id to user id table. Also, this means that in the doesUserIdExist query in generalqueries, we do not need to do a UNION. - UNION is not required as we also have a fkey constraint, appidtouserid table is a single source for truth.
- [ ] Optimise the getPrimaryUserInfoForUserIds function to not query all auth recipes all the time, and instead, do it based on the recipe ids of the fetched user.
- [x] Test many signIn using 500 threads in parallel and see the time difference between before accountlinking and after.
- [x] When linking accounts, remove active user for user that was linked and add active user with primaryUserId
- [x] Test that `primary_or_recipe_user_time_joined` is updated correctly during account linking and unlinking
- [x] Change in chagelog about `primary_or_recipe_user_time_joined`, index related to that and the migration script
- [x] paid feature stats
- [x] test user search with account linking
- [x] Update CDI about returning recipeUserId in sign in/up related APIs and user object in link accounts API
- [x] uncomment assert on isSetExternalUserIdCalled in LoginMethod class
- [x] Update thirdparty sign in up API in CDI to accept `isVerified`
- [x] see where all listUser**_Transaction are being called and try to minimize the code paths to those functions.
- [x] paid feature stats -> send boolean true if even one of the users is a primary user + send number of active users that have account linking in them + number of total users that have account linking in them
- [ ] in the test for testNetworkCallIsMadeInCoreInit, does it test with all the features enabled? For examlpe, if multi tenanct was not enabled, then the bug would not have been caught.. but what is stopping this bug to come again with other features like account linking or totp or anything else? you have only added the license key feature for multi tenancy here.. please change the test to account for all EE features and also such that if we add new EE features, they should be accounted for automatically, or the test should fail if the it’s not modified.
- [x] the get user api for EP returns non EP users as well


rishabhpoddar:
## DB PRs (for sqllite):
- linkAccount function
- https://github.com/supertokens/supertokens-postgresql-plugin/pull/136

rishabhpoddar:
## Test cases
- [ ] CreatePrimaryUser
   - [ ] locking test - make several requests for two users with the same email with making them primary and not primary and then check that the db state is always consistent.
   - [ ] locking test - make sure that deadlocks are resolved on their own.
   - [ ] check for making primary user across tenants logic.
- [ ] get user by id transaction (and without transaction) storaget layer test
   - [ ] test with more than 4 login methods for one user, across all three auth recipes

PR #830, opened 2023-10-07
normalized input of parseStringOrThrowError method

### LINKED ISSUE #594: `InputParser.parseStringOrThrowError` should additionally normalise the input
## 🐛 Bug Report

(A clear and concise description of what the bug is)

## Useful informations

(Write what happened. Add screenshots, stacktraces, videos, anything that can help)

Taz03:
## Related issues

fixes #594 

    - In `build.gradle`
- [x] If there are new dependencies that have been added in `build.gradle`, please make sure to add them
- [x] Update function `getValidFields` in `io/supertokens/config/CoreConfig.java` if new aliases were added for any core config (similar to the `access_token_signing_key_update_interval` config alias).
- [x] If added a foreign key constraint on `app_id_to_user_id` table, make sure to delete from this table when deleting the user as well if `deleteUserIdMappingToo` is false.


src/main/java/io/supertokens/webserver/InputParser.java
```diff
@@ -137,7 +137,11 @@ public static String parseStringOrThrowError(JsonObject element, String fieldNam
             if (!stringified.contains("\"")) {
                 throw new Exception();
             }
-            return ((JsonObject) element).get(fieldName).getAsString();
+            String s = element.get(fieldName).getAsString().trim();
+            if (s.contains("@")) {
+                s = s.toLowerCase();
+            }
```
rishabhpoddar, line 143: We can't assume that if there is an `@`, it must be for an email input. Instead, we should use the full email regex to check, and if it passes that, only then we should do the toLowerCase.

src/main/java/io/supertokens/webserver/InputParser.java
```diff
@@ -137,7 +137,11 @@ public static String parseStringOrThrowError(JsonObject element, String fieldNam
...
+            return s;
         } catch (Exception e) {
```
rishabhpoddar, line 145: Also, please add tests for this change please :)

rishabhpoddar (changes requested):

src/main/java/io/supertokens/webserver/InputParser.java
Taz03: Done

src/main/java/io/supertokens/webserver/InputParser.java
Taz03: Done

src/main/java/io/supertokens/webserver/InputParser.java
```diff
@@ -137,7 +137,12 @@ public static String parseStringOrThrowError(JsonObject element, String fieldNam
             if (!stringified.contains("\"")) {
                 throw new Exception();
             }
-            return ((JsonObject) element).get(fieldName).getAsString();
+            String s = element.get(fieldName).getAsString().trim();
+
+            if (s.matches("^(?=.{1,64}@)[A-Za-z0-9_-]+(\\.[A-Za-z0-9_-]+)*@[^-][A-Za-z0-9-]+(\\.[A-Za-z0-9-]+)*(\\.[A-Za-z]{2,})$")) {
```
rishabhpoddar, line 142: We use the following regex in our backend SDK: https://github.com/supertokens/supertokens-node/blob/master/lib/ts/recipe/emailpassword/utils.ts#L242. 

Might want to make this consistent with that.

rishabhpoddar (changes requested):

src/main/java/io/supertokens/webserver/InputParser.java
```diff
@@ -137,7 +137,12 @@ public static String parseStringOrThrowError(JsonObject element, String fieldNam
             if (!stringified.contains("\"")) {
                 throw new Exception();
             }
-            return ((JsonObject) element).get(fieldName).getAsString();
```
rishabhpoddar, line 140: This function is called for non GET request. We might want to add the trim and toLowerCase (in case of email), to the GET request equivalent as well. The functions are `getQueryParamOrThrowError` and `getCommaSeparatedStringArrayQueryParamOrThrowError`

rishabhpoddar (changes requested):

src/main/java/io/supertokens/webserver/InputParser.java
Taz03: do we want to trim in case of non-email in those methods as well?

src/main/java/io/supertokens/webserver/InputParser.java
Taz03: Done

src/main/java/io/supertokens/webserver/InputParser.java
Taz03: Done

PR #832, opened 2023-10-11
fix: duplicate cron task

sattvikc:
## Summary of change

Task shouldn't be scheduled to executor if the task already exists in the tasks list.

## Test Plan

Manually tested that right number of API calls is being done. Couldn't find a way to count number of scheduled executors.

## Documentation changes

    - In `build.gradle`

- [ ] Item1
- [ ] Item2

### DIFF
--- a/src/main/java/io/supertokens/cronjobs/Cronjobs.java
+++ b/src/main/java/io/supertokens/cronjobs/Cronjobs.java
@@ -91,9 +91,9 @@ public class Cronjobs extends ResourceDistributor.SingletonResource {
         }
         Cronjobs instance = getInstance(main);
         synchronized (instance.lock) {
-            instance.executor.scheduleWithFixedDelay(task, task.getInitialWaitTimeSeconds(),
-                    task.getIntervalTimeSeconds(), TimeUnit.SECONDS);
             if (!instance.tasks.contains(task)) {
+                instance.executor.scheduleWithFixedDelay(task, task.getInitialWaitTimeSeconds(),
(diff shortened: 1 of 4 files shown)


src/main/java/io/supertokens/cronjobs/Cronjobs.java
```diff
@@ -91,9 +91,9 @@ public static void addCronjob(Main main, CronTask task) {
         }
         Cronjobs instance = getInstance(main);
         synchronized (instance.lock) {
-            instance.executor.scheduleWithFixedDelay(task, task.getInitialWaitTimeSeconds(),
-                    task.getIntervalTimeSeconds(), TimeUnit.SECONDS);
             if (!instance.tasks.contains(task)) {
+                instance.executor.scheduleWithFixedDelay(task, task.getInitialWaitTimeSeconds(),
+                        task.getIntervalTimeSeconds(), TimeUnit.SECONDS);
                 instance.tasks.add(task);
```
rishabhpoddar, line 97: add test for this

rishabhpoddar (changes requested):

src/main/java/io/supertokens/cronjobs/Cronjobs.java
rishabhpoddar: And why does the addCronjob get called so many times for the same job anyway?

rishabhpoddar: Also, it's possible that the task obj is a different instance of the same class?

rishabhpoddar (changes requested):

Issue #880, opened 2023-11-15
Phone Number Normalization Discrepancy in SDKs and APIs


anku255:
## 🐛 Bug Report

Currently, phone number normalization relies on SDKs, leading to a potential disparity between the queried phoneNumber and the actual stored value in the database. Consider the following scenario: 

```
input phoneNumber -> 555-123-4567

The createUser function in the SDK doesn't normalize this input, storing it as-is in the Database.

input phoneNumber -> 555-123-4567

searchUserByPhoneNumber normalizes the input to +5551234567, but it is unable to locate the user as the phone number doesn't match.

```

## Suggestion Solution

Implement phone number normalization at the core level for every API. This approach would ensure consistent behavior across all APIs and SDKs.

## Backwards Compatibility

To maintain compatibility, we propose adding a cron job to normalize all stored phone numbers in the database. This cron job would also introduce a flag, such as `isPhoneNumberNormalized`, ensuring that phone numbers are normalized only once.

The default flag value for any new user would be true since the APIs would have already normalized it during the initial processing.

## TODO:
- [x] add test where input is not a phone number
- [x] migration script
- [ ] update saas infra migration code to run the custom script

rishabhpoddar:
This has been fixed in core version >= 7.0.12. For users migrating to this and using the passwordless recipe, please be sure to checkout [the changelog](https://github.com/supertokens/supertokens-core/blob/master/CHANGELOG.md), for migration instructions that normalises existing phone numbers in the database.

PR #881, opened 2023-11-16
feat: Normalise phoneNumber received in the input before processing

anku255:
## Summary of change

This pull request introduces phone number normalisation to all API endpoints that accept phone numbers as user input. The implementation includes the addition of a `normaliseIfPhoneNumber` function in the `Utils` class, which is now utilized across all relevant APIs. 

## Related issues

https://github.com/supertokens/supertokens-core/issues/880

## Test Plan

I've added test cases to ensure the proper normalization of phone numbers by the `normaliseIfPhoneNumber` utility function. Additionally, I've added corresponding test cases in all modified APIs to validate this behavior. 

## Documentation changes

(If relevant, please create a PR in our [docs repo](https://github.com/supertokens/docs), or create a checklist here
highlighting the necessary changes)

## Checklist for important updates

- [ ] Changelog has been updated
    - [ ] If there are any db schema changes, mention those changes clearly
- [ ] `coreDriverInterfaceSupported.json` file has been updated (if needed)
- [ ] `pluginInterfaceSupported.json` file has been updated (if needed)
- [ ] Changes to the version if needed
    - In `build.gradle`
- [ ] If added a new paid feature, edit the `getPaidFeatureStats` function in FeatureFlag.java file
- [x] Had installed and ran the pre-commit hook
- [x] If there are new dependencies that have been added in `build.gradle`, please make sure to add them
  in `implementationDependencies.json`.
- [ ] Update function `getValidFields` in `io/supertokens/config/CoreConfig.java` if new aliases were added for any core config (similar to the `access_token_signing_key_update_interval` config alias).
- [x] Issue this PR against the latest non released version branch.
    - To know which one it is, run find the latest released tag (`git tag`) in the format `vX.Y.Z`, and then find the
      latest branch (`git branch --all`) whose `X.Y` is greater than the latest released tag.
    - If no such branch exists, then create one from the latest released branch.
- [ ] If added a foreign key constraint on `app_id_to_user_id` table, make sure to delete from this table when deleting the user as well if `deleteUserIdMappingToo` is false.

PR #882, opened 2023-11-17
feat: Add migration script for phone number normalisation

anku255:
## Summary of change

This pull request introduces phone number normalisation migration script that is supposed to be run on the all the core instances after migrating to version 7.0.12.

## Related issues

- https://github.com/supertokens/supertokens-core/issues/880

## Test Plan

I locally tested the script by creating `passwordless_users`, `passwordless_devices`, and `passwordless_user_to_tenant` tables, seeding them with unnormalized phone numbers, and confirming successful normalization after script execution. Tests covered both PostgreSQL and MySQL databases.

## Screenshots

[Click here](https://app.warp.dev/block/TdJNIlLCx8LtCsBMpAkrG6) to see the script output logs when run with 20K rows in each table.

    - In `build.gradle`


CHANGELOG.md
```diff
@@ -7,7 +7,26 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [7.0.12] - 2023-11-16
 
-- Adds Phone Number normalisation
+In this release, the core API routes have been updated to incorporate phone number normalization before processing. Consequently, existing entries in the database also need to undergo normalization. To facilitate this, we have included a migration script to normalize phone numbers for all the existing entries.
```
anku255, line 10: @rishabhpoddar Wasn't sure how this migration script will be executed. Expecting some changes here.

migration_scripts/to_version_7_1/index.js
```diff
@@ -0,0 +1,118 @@
...
+  }
+}
+
+async function updatePhoneNumbers(table) {
+  const batchSize = 1000;
+  let offset = 0;
+
+  try {
+    while (true) {
+      const entries = await knex.raw(`SELECT * FROM ${table} WHERE phone_number is NOT NULL LIMIT ${batchSize} OFFSET ${offset}`);
```
anku255, line 76: @rishabhpoddar There is where clause for `phone_number is NOT NULL`. Let me know if you want me to remove that as this field may not be indexed in all the tables.

This should be problem only if the tables are supposed to be massive.

migration_scripts/to_version_7_1/index.js
```diff
@@ -0,0 +1,118 @@
...
+      const rows = entries.rows ? entries.rows : entries[0];
+
+      const batchUpdates = [];
+
+      for (const entry of rows) {
+        const currentPhoneNumber = entry.phone_number;
+        const normalizedPhoneNumber = getNormalizedPhoneNumber(currentPhoneNumber);
+
+        if (normalizedPhoneNumber && normalizedPhoneNumber !== currentPhoneNumber) {
+          const updateQuery = getUpdateQuery(table, entry, normalizedPhoneNumber);
```
rishabhpoddar, line 86: One constraint to consider is that `passwordless_users` and `passwordless_user_to_tenant` are supposed to be consistent. So when updating `passwordless_users`, for a user_id and an app_id, we should also update `passwordless_user_to_tenant` for that user_id and app_id within a transaction.

This also means that you don't need to loop through `passwordless_user_to_tenant`.

rishabhpoddar (changes requested):

migration_scripts/to_version_7_1/index.js
```diff
@@ -0,0 +1,118 @@
...
+const DB_PASSWORD = "";
+const DB_NAME = "";
+const CLIENT = ""; // Use "pg" for PostgreSQL and "mysql2" for MySQL DB
+
+if(!DB_HOST || !CLIENT) {
+  console.error('Please update the DB_HOST, DB_USER, DB_PASSWORD, DB_DATABASE and CLIENT variables before running the script.');
+  return;
+}
+
+const knex = require('knex')({
```
rishabhpoddar, line 31: how many db connections did this script use? I assume just one?

rishabhpoddar (changes requested):

migration_scripts/to_version_7_1/index.js
anku255: done

CHANGELOG.md
```diff
@@ -7,7 +7,26 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
...
+1. Ensure that the core is already upgraded to version 7.0.12 (CDI version 4.0)
+2. Run the migration script
+
+    <details>
+    <summary>Steps</summary>
+
+    Make sure your Node.js version is 16 or above to run the script. Locate the migration script at `supertokens-core/migration_scripts/to_version_7_1/index.js`. Modify the script by updating the `DB_HOST`, `DB_USER`, `DB_PASSWORD`, and `DB_NAME` variables with the correct values. Subsequently, run the following commands to initiate the script:
+    
+    ```bash
+       $ cd supertokens-core/migration_scripts/to_version_7_1
```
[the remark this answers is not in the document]
anku255: Done (build.gradle changes were already done in #881)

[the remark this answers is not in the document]
anku255: done

CHANGELOG.md
```diff
@@ -7,7 +7,26 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
...
+
+**NOTE**: You can skip the migration if you are not using passwordless via phone number.
+
+### Migration steps
+
+1. Ensure that the core is already upgraded to version 7.0.12 (CDI version 4.0)
+2. Run the migration script
+
+    <details>
+    <summary>Steps</summary>
```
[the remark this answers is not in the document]
anku255: done

migration_scripts/to_version_7_1/index.js
anku255: Yeah few 100K should be absolutely fine.

migration_scripts/to_version_7_1/index.js
anku255: Yeah it uses just 1. The default pool config was { min: 2, max: 10 }. I have changed that to a more conservative value of  { min: 0, max: 5 }.

CHANGELOG.md
```diff
@@ -7,7 +7,23 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
...
+In this release, the core API routes have been updated to incorporate phone number normalization before processing. Consequently, existing entries in the database also need to undergo normalization. To facilitate this, we have included a migration script to normalize phone numbers for all the existing entries.
+
+**NOTE**: You can skip the migration if you are not using passwordless via phone number.
+
+### Migration steps
+
+1. Ensure that the core is already upgraded to version 7.0.12 (CDI version 4.0)
+2. Run the migration script
+
+    Make sure your Node.js version is 16 or above to run the script. Locate the migration script at `supertokens-core/migration_scripts/to_version_7_1_12/index.js`. Modify the script by updating the `DB_HOST`, `DB_USER`, `DB_PASSWORD`, and `DB_NAME` variables with the correct values. Subsequently, run the following commands to initiate the script:
```
rishabhpoddar, line 19:
```suggestion
    Make sure your Node.js version is 16 or above to run the script. Locate the migration script at `supertokens-core/migration_scripts/to_version_7_0_12/index.js`. Modify the script by updating the `DB_HOST`, `DB_USER`, `DB_PASSWORD`, and `DB_NAME` variables with the correct values. Subsequently, run the following commands to initiate the script:
```

rishabhpoddar: Folder name should be changed to 7_0_12 cause the version being released is 7.0.12

CHANGELOG.md
```diff
@@ -7,7 +7,23 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [7.0.12] - 2023-11-16
 
-- Adds Phone Number normalisation
+In this release, the core API routes have been updated to incorporate phone number normalization before processing. Consequently, existing entries in the database also need to undergo normalization. To facilitate this, we have included a migration script to normalize phone numbers for all the existing entries.
+
+**NOTE**: You can skip the migration if you are not using passwordless via phone number.
+
+### Migration steps
+
```
rishabhpoddar, line 15: Write what this migration script is doing. Something like "It normalises existing phone numbers in the database if they are unnormalised"

migration_scripts/to_version_7_1/index.js
rishabhpoddar: We should try and use minimal number of connections since it's happening in a live db. Can we just set it to min: 1, max: 1 so that it uses just one connection for the whole script?

migration_scripts/to_version_7_1_12/index.js
```diff
@@ -0,0 +1,110 @@
...
+}
+
+async function updatePhoneNumbers(table) {
+  const batchSize = 1000;
+  let offset = 0;
+
+  try {
+    while (true) {
+      const entries = await knex.raw(`SELECT * FROM ${table} WHERE phone_number is NOT NULL LIMIT ${batchSize} OFFSET ${offset}`);
+      const rows = entries.rows ? entries.rows : entries[0];
```
rishabhpoddar, line 69: what does entries[0] point to? I assume `entries.rows` will the `false` if there are no passwordless users?

migration_scripts/to_version_7_1_12/index.js
```diff
@@ -0,0 +1,110 @@
...
+        break;
+      }
+    }
+  } catch (error) {
+    console.error(`Error normalising phone numbers for table ${table}:`, error.message);
+  }
+}
+
+async function runScript() {
+  const tables = ['passwordless_users', 'passwordless_devices', 'passwordless_user_to_tenant'];
```
rishabhpoddar, line 99:
```suggestion
  const tables = ['passwordless_users', 'passwordless_devices'];
```

rishabhpoddar (changes requested):

migration_scripts/to_version_7_1/index.js
rishabhpoddar: Whats the max number of rows that you have tested it with for mysql and postgres?

migration_scripts/to_version_7_1_12/index.js
```diff
@@ -0,0 +1,110 @@
...
+      offset += rows.length;
+
+      console.log(`Processed ${offset} rows for table ${table}`);
+
+      if (rows.length < batchSize) {
+        break;
+      }
+    }
+  } catch (error) {
+    console.error(`Error normalising phone numbers for table ${table}:`, error.message);
```
rishabhpoddar, line 94: The execution will come here anytime any query will fail right (even if a single update query fails for example). In this case, we should just ask the user to rerun the whole script? If so, please change the error message to contain that too.

migration_scripts/to_version_7_1_12/index.js
```diff
@@ -0,0 +1,110 @@
...
+const DB_PASSWORD = "";
+const DB_NAME = "";
+const CLIENT = ""; // Use "pg" for PostgreSQL and "mysql2" for MySQL DB
+
+if (!DB_HOST || !CLIENT) {
+  console.error('Please update the DB_HOST, DB_USER, DB_PASSWORD, DB_DATABASE and CLIENT variables before running the script.');
+  return;
+}
+
+const knex = require('knex')({
```
rishabhpoddar, line 31: whats the default connection and query timeout for this? We should keep them large (min 10 seconds for both)

migration_scripts/to_version_7_1_12/index.js
```diff
@@ -0,0 +1,110 @@
...
+          const updatePromise = getUpdatePromise(table, entry, normalizedPhoneNumber);
+          batchUpdates.push(updatePromise);
+        }
+      }
+
+      await Promise.all(batchUpdates);
+
+      offset += rows.length;
+
+      console.log(`Processed ${offset} rows for table ${table}`);
```
rishabhpoddar, line 87: Would be nice to print out how many rows in total as well.

rishabhpoddar (changes requested):

migration_scripts/to_version_7_1/index.js
anku255: Updating `passwordless_devices` table is really fast regardless of connection pool size. But for `passwordless_users` and `passworldless_user_to_tenant` tables the script on average takes - 

```
4500ms per 1000 rows with MAX 10 connections
4700ms per 1000 rows with MAX 5 connections
19000ms per 1000 rows with MAX 1 connections
```

Seeing this, I believe the default `MAX_POOL_SIZE` should be 5. I have also updated the changelog informing user that they can increase this value to potentially increase the speed even more.

migration_scripts/to_version_7_1/index.js
anku255: I have tested it with 200K rows. The speed depends on the number of connection as explained [here](https://github.com/supertokens/supertokens-core/pull/882#discussion_r1398799600)

CHANGELOG.md
anku255: Thanks for catching this. Somehow, 7.0.12 changed to 7.1.12 in my brain.

CHANGELOG.md
anku255: done

migration_scripts/to_version_7_1_12/index.js
anku255: This is because knex is using `pg` and `mysql2` lib underneath the hood and they have different return signatures. I have added a comment to clarify this.

migration_scripts/to_version_7_1_12/index.js
anku255: done.  I added `totalUpdatedRows` as the we updated the row only if it changes after normalisation.

migration_scripts/to_version_7_1_12/index.js
anku255: default `acquireConnectionTimeout` is `60000ms`

anku255: There is no timeout for query but based on the suggestion I have also added a  `QUERY_TIMEOUT` of 60s.

migration_scripts/to_version_7_1_12/index.js
anku255: Done!

CHANGELOG.md
```diff
@@ -7,7 +7,27 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
...
+    Make sure your Node.js version is 16 or above to run the script. Locate the migration script at `supertokens-core/migration_scripts/to_version_7_0_12/index.js`. Modify the script by updating the `DB_HOST`, `DB_USER`, `DB_PASSWORD`, and `DB_NAME` variables with the correct values. Subsequently, run the following commands to initiate the script:
+    
+    ```bash
+       $ git clone https://github.com/supertokens/supertokens-core.git
+       $ cd supertokens-core/migration_scripts/to_version_7_0_12
+       $ npm install
+       $ npm start
+    ```
+
+    NOTE: If the script is experiencing slow performance, consider adjusting the `MAX_POOL_SIZE` to a higher value. Increasing this parameter allows the script to leverage more connections simultaneously, potentially improving execution speed.
```
rishabhpoddar, line 30: You  should add the timing values from this comment here too: https://github.com/supertokens/supertokens-core/pull/882#discussion_r1398799600

migration_scripts/to_version_7_1_12/index.js
rishabhpoddar: I was thinking more like how many rows in total, and not just for the current iteration. Sort of like "Processed X/Y rows" where X is `offset` and Y is the total number of rows. This is so that the dev gets a sense of how much time is pending.

Also, please attach screenshots of the script output in this PR description with all the various console logs showing.

migration_scripts/to_version_7_1/index.js
rishabhpoddar: Cool

migration_scripts/to_version_7_1_12/index.js
anku255: Ah got it! that makes sense. I have added it now.

CHANGELOG.md
anku255: done!

migration_scripts/to_version_7_0_12/index.js
```diff
@@ -70,10 +68,15 @@ function getNormalizedPhoneNumber(phoneNumber) {
 async function updatePhoneNumbers(table) {
   const batchSize = 1000;
   let offset = 0;
+  let totalUpdatedRows = 0;
 
   try {
     while (true) {
+      let totalRows = await knex.raw(`SELECT COUNT(*) as count FROM ${table} WHERE phone_number is NOT NULL`);
```
rishabhpoddar, line 75: this query should not be a part of the loop.

rishabhpoddar (changes requested):

migration_scripts/to_version_7_0_12/index.js
anku255: Thank you so much for catching that. That was really silly. Unfortunately. there was nothing in logs that could point this out. 

I have fixed it now.

rishabhpoddar (approved):

PR #948, opened 2024-03-08
fix: add check code API and update delete code API

sattvikc:
## Summary of change

## Related issues

- Link to issue1 here
- Link to issue1 here

## Test Plan

## Documentation changes

    - In `build.gradle`

- [ ] Item1
- [ ] Item2

### DIFF
--- a/src/main/java/io/supertokens/passwordless/Passwordless.java
+++ b/src/main/java/io/supertokens/passwordless/Passwordless.java
@@ -581,8 +581,8 @@ public class Passwordless {
         });
     }
 
-    public static void removeCodeByDeviceIdHash(TenantIdentifier tenantIdentifier, Storage storage,
-                                                String deviceIdHash)
+    public static void removeDevice(TenantIdentifier tenantIdentifier, Storage storage,
+                                    String deviceIdHash)
             throws StorageQueryException, StorageTransactionLogicException {
         PasswordlessSQLStorage passwordlessStorage = StorageUtils.getPasswordlessStorage(storage);
 
--- a/src/main/java/io/supertokens/passwordless/Passwordless.java
+++ b/src/main/java/io/supertokens/passwordless/Passwordless.java
@@ -287,9 +287,9 @@ public class Passwordless {
                 false);
     }
 
-    public static PasswordlessDevice verifyCode(TenantIdentifier tenantIdentifier, Storage storage, Main main,
-                                                String deviceId, String deviceIdHashFromUser,
-                                                String userInputCode, String linkCode)
+    public static PasswordlessDevice checkCodeAndReturnDevice(TenantIdentifier tenantIdentifier, Storage storage, Main main,
+                                                              String deviceId, String deviceIdHashFromUser,
+                                                              String userInputCode, String linkCode, boolean deleteCodeOnSuccess)
             throws RestartFlowException, ExpiredUserInputCodeException,
             IncorrectUserInputCodeException, DeviceIdHashMismatchException, StorageTransactionLogicException,
             StorageQueryException, NoSuchAlgorithmException, InvalidKeyException, IOException, Base64EncodingException,
@@ -385,12 +385,14 @@ public class Passwordless {
                     throw new StorageTransactionLogicException(new RestartFlowException());
                 }
 
-                if (device.email != null) {
-                    passwordlessStorage.deleteDevicesByEmail_Transaction(tenantIdentifier, con,
-                            device.email);
-                } else if (device.phoneNumber != null) {
-                    passwordlessStorage.deleteDevicesByPhoneNumber_Transaction(tenantIdentifier, con,
-                            device.phoneNumber);
+                if (deleteCodeOnSuccess) {
+                    if (device.email != null) {
+                        passwordlessStorage.deleteDevicesByEmail_Transaction(tenantIdentifier, con,
+                                device.email);
+                    } else if (device.phoneNumber != null) {
+                        passwordlessStorage.deleteDevicesByPhoneNumber_Transaction(tenantIdentifier, con,
+                                device.phoneNumber);
+                    }
                 }
 
                 passwordlessStorage.commitTransaction(con);
@@ -420,8 +422,8 @@ public class Passwordless {
 
         PasswordlessSQLStorage passwordlessStorage = StorageUtils.getPasswordlessStorage(storage);
 
-        PasswordlessDevice consumedDevice = verifyCode(tenantIdentifier, storage, main, deviceId, deviceIdHashFromUser,
-                userInputCode, linkCode);
+        PasswordlessDevice consumedDevice = checkCodeAndReturnDevice(tenantIdentifier, storage, main, deviceId, deviceIdHashFromUser,
+                userInputCode, linkCode, true);
 
         // Getting here means that we successfully consumed the code
         AuthRecipeUserInfo user = null;
--- a/src/main/java/io/supertokens/passwordless/Passwordless.java
+++ b/src/main/java/io/supertokens/passwordless/Passwordless.java
@@ -252,7 +252,7 @@ public class Passwordless {
             Storage storage = StorageLayer.getStorage(main);
             return consumeCode(
                     new TenantIdentifier(null, null, null), storage,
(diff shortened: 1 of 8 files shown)


src/main/java/io/supertokens/passwordless/Passwordless.java
```diff
@@ -440,58 +440,84 @@ public static ConsumeCodeResponse consumeCode(TenantIdentifier tenantIdentifier,
...
+        if (user != null) {
+            // We do not need this cleanup if we are creating the user, since it uses the email/phoneNumber of the
+            // device, which has already been cleaned up
+            if (loginMethod.email != null && !loginMethod.email.equals(consumedDevice.email)) {
+                removeCodesByEmail(tenantIdentifier, storage, loginMethod.email);
+            }
+            if (loginMethod.phoneNumber != null && !loginMethod.phoneNumber.equals(consumedDevice.phoneNumber)) {
+                removeCodesByPhoneNumber(tenantIdentifier, storage, loginMethod.phoneNumber);
+            }
+        }
```
rishabhpoddar, line 452: why are we doing this?

src/main/java/io/supertokens/webserver/api/passwordless/VerifyCodeAPI.java
```diff
@@ -0,0 +1,146 @@
...
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws IOException, ServletException {
+        // API is tenant specific
+        // Logic based on: https://app.code2flow.com/OFxcbh1FNLXd
+        JsonObject input = InputParser.parseJsonObjectOrThrowError(req);
+
+        String linkCode = null;
+        String deviceId = null;
+        String userInputCode = null;
+
```
rishabhpoddar, line 64: there is also a deleteCode boolean input.

src/main/java/io/supertokens/passwordless/Passwordless.java
```diff
@@ -440,58 +440,84 @@ public static ConsumeCodeResponse consumeCode(TenantIdentifier tenantIdentifier,
             }
         }
 
+        if (user != null) {
+            // We do not need this cleanup if we are creating the user, since it uses the email/phoneNumber of the
+            // device, which has already been cleaned up
+            if (loginMethod.email != null && !loginMethod.email.equals(consumedDevice.email)) {
```
rishabhpoddar, line 446: u need to also do the email verification stuff here?

src/main/java/io/supertokens/passwordless/Passwordless.java
```diff
@@ -440,58 +440,84 @@ public static ConsumeCodeResponse consumeCode(TenantIdentifier tenantIdentifier,
...
+                    // by cleaning up the device/code this consume would've used.
+                    throw new RestartFlowException();
+                } catch (DuplicateUserIdException e) {
+                    // We can retry..
                 }
             }
         } else {
-            // We do not need this cleanup if we are creating the user, since it uses the email/phoneNumber of the
-            // device, which has already been cleaned up
             if (setEmailVerified && consumedDevice.email != null) {
```
rishabhpoddar, line 521: the suff in the else branch needs to go away since it will be done in verifyCode

src/main/java/io/supertokens/passwordless/Passwordless.java
```diff
@@ -440,58 +440,84 @@ public static ConsumeCodeResponse consumeCode(TenantIdentifier tenantIdentifier,
             }
         }
 
+        if (user != null) {
```
rishabhpoddar, line 443: in this function, you need to only delete the code if deleteCode is true. Else, you need to do everything else you are doing, except for actually deleting the code

src/main/java/io/supertokens/webserver/Webserver.java
```diff
@@ -187,6 +187,7 @@ private void setupRoutes() {
         addAPI(new DeleteCodesAPI(main));
         addAPI(new DeleteCodeAPI(main));
         addAPI(new CreateCodeAPI(main));
+        addAPI(new VerifyCodeAPI(main));
```
rishabhpoddar, line 190: there is no new api. The existing consume code api is changed

src/main/java/io/supertokens/webserver/api/passwordless/ConsumeCodeAPI.java
```diff
@@ -65,7 +65,6 @@ protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws I
         String linkCode = null;
         String deviceId = null;
         String userInputCode = null;
-        Boolean createRecipeUserIfNotExists = true;
```
rishabhpoddar, line 68: this boolean is still there i think

rishabhpoddar (changes requested):

porcellus (changes requested):

src/main/java/io/supertokens/passwordless/Passwordless.java
porcellus: I don't think this is applicable anymore.

porcellus (approved):

src/main/java/io/supertokens/passwordless/Passwordless.java
```diff
@@ -518,6 +529,8 @@ public static ConsumeCodeResponse consumeCode(TenantIdentifier tenantIdentifier,
                 }
             }
 
+            // We do not need this cleanup if we are creating the user, since it uses the email/phoneNumber of the
+            // device, which has already been cleaned up
```
rishabhpoddar, line 533: this comment says we do not need this cleanuip, but still there is some cleanup happening below. So it's confusing. Also, why don't we need the cleanup? Here we are deleting the other codes that belong to the email / phone number

src/main/java/io/supertokens/passwordless/Passwordless.java
```diff
@@ -409,6 +410,20 @@ public static ConsumeCodeResponse consumeCode(TenantIdentifier tenantIdentifier,
             }
             throw e;
         }
+    }
+
+    public static ConsumeCodeResponse consumeCode(TenantIdentifier tenantIdentifier, Storage storage, Main main,
```
rishabhpoddar, line 415: the diff makes it hard to see, but, is the only change in this function only that you are calling checkCodeAndReturnDevice at the start instead of doing the code consume transaction in this function itself? What happens after that remains the same right?

src/main/java/io/supertokens/passwordless/Passwordless.java
```diff
@@ -570,6 +583,18 @@ public static void removeCode(TenantIdentifier tenantIdentifier, Storage storage
         });
     }
 
+    public static void removeDevice(TenantIdentifier tenantIdentifier, Storage storage,
+                                    String deviceIdHash)
+            throws StorageQueryException, StorageTransactionLogicException {
+        PasswordlessSQLStorage passwordlessStorage = StorageUtils.getPasswordlessStorage(storage);
+
+        passwordlessStorage.startTransaction(con -> {
+            passwordlessStorage.deleteDevice_Transaction(tenantIdentifier, con, deviceIdHash);
```
rishabhpoddar, line 592: this will delete the passwordless code and everything related to that device right? For example, if there are many codes created for the device (is that even possible?), it will delete all of them right?

rishabhpoddar (changes requested):

src/main/java/io/supertokens/passwordless/Passwordless.java
sattvikc: yes, they get deleted via FK cascade delete

src/main/java/io/supertokens/passwordless/Passwordless.java
sattvikc: Technically there is no code change in the consume code function. First part where the code check was happening is refactored into a separate function. only change is that we pass a flag to decide whether we want to delete the device on success or not

src/main/java/io/supertokens/passwordless/Passwordless.java
sattvikc: this is an old comment, I just moved it to the relevant place. anyway, it is talking about the other branch in the if that does not need clean up, I have rephrased it to make it clear.

PR #954, opened 2024-03-12
feat: merge last active times when linking users

porcellus:
## Summary of change

merge last active times when linking users

## Related issues

- 

## Test Plan

Added test cases that check that the users are merged and get the latest activity time

## Documentation changes

N/A

    - In `build.gradle`

- [ ] Item1
- [ ] Item2

### DIFF
--- a/src/main/java/io/supertokens/authRecipe/AuthRecipe.java
+++ b/src/main/java/io/supertokens/authRecipe/AuthRecipe.java
@@ -16,6 +16,7 @@
 
 package io.supertokens.authRecipe;
 
+import io.supertokens.ActiveUsers;
 import io.supertokens.Main;
 import io.supertokens.authRecipe.exception.AccountInfoAlreadyAssociatedWithAnotherPrimaryUserIdException;
 import io.supertokens.authRecipe.exception.InputUserIdIsNotAPrimaryUserException;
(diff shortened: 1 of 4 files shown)


src/main/java/io/supertokens/authRecipe/AuthRecipe.java
```diff
@@ -348,10 +347,19 @@ public static LinkAccountsResult linkAccounts(Main main, AppIdentifier appIdenti
...
                             canLinkAccounts.primaryUserId);
 
+                    long recipeUserLastActive = activeUsersStorage.getLastActiveByUserId_Transaction(con, appIdentifier, canLinkAccounts.recipeUserId);
+                    long primaryUserLastActive = activeUsersStorage.getLastActiveByUserId_Transaction(con, appIdentifier, canLinkAccounts.primaryUserId);
+
+                    if (primaryUserLastActive < recipeUserLastActive) {
+                        activeUsersStorage.setUserActive_Transaction(con, appIdentifier, canLinkAccounts.primaryUserId, recipeUserLastActive);
+                    }
+                    activeUsersStorage.deleteUserActive_Transaction(con, appIdentifier, canLinkAccounts.recipeUserId);
+
```
rishabhpoddar, line 362: - what happens during unlink in regards to active users?
- Ideally, this stuff should happen outside the scope of the actual account linking. So after the linkAccounts function has successfully returned, in the API, we should do this update in a separate transaction. This is so that even if this transaction fails, the API should still return a 200.
- In this, we can just set the primary user's last active time to now, and delete the entry for the recipeuserid (if it's different from the primary user id). No need to fetch the values first. This way, it's 2 less queries. In fact, both of these queires can happen on their own, no need for a transaction either. So ideally, the only change required to the plugin interface is to add a deleteUserActive function.

rishabhpoddar (changes requested):

src/main/java/io/supertokens/authRecipe/AuthRecipe.java
porcellus: - Nothing, and I think this makes sense: if the user was unlinked, the last activity time should remain unchanged, and if the user was deleted it should still be included in the monthly active users calculation.
- Sure, I figured all these ops were worth doing in a single transaction but I guess separating it out makes more sense.
- I tried to do merging so we do not mark all users as active in a mass migration scenario, but I guess this is also fine. Should I make createPrimaryUser also mark the user as active? I don't think that happens right now.

src/main/java/io/supertokens/authRecipe/AuthRecipe.java
```diff
@@ -362,6 +367,15 @@ public static LinkAccountsResult linkAccounts(Main main, AppIdentifier appIdenti
                 }
             });
 
+            if (!result.wasAlreadyLinked) {
+                try {
+                    ActiveUsers.updateLastActiveAfterLinking(
```
rishabhpoddar, line 372: move to api layer

rishabhpoddar (changes requested):

rishabhpoddar (approved):

Issue #38, opened 2020-09-28
Make logging on start prettier


rishabhpoddar:

rahullli:
I would like to work on this issue .

rishabhpoddar:
@rahullli  please see the updated first comment.

PR #974, opened 2024-04-02
logging on start prettier

nikhilbhatia08:
## Summary of change

Made logging on start prettier by adding a superTokens starter template and added color to the output 
## Related issues

- [Link to issue1 here](https://github.com/supertokens/supertokens-core/issues/38)

## Test Plan

When the core is run then it gives output with colors and the starter template and all the tests have been passed
Screenshot 2024-04-02 at 12 29 49 PM
This is the picture of the output with colors and the starter template

No remaining TODO's for this PR

### DIFF
--- a/src/main/java/io/supertokens/Main.java
+++ b/src/main/java/io/supertokens/Main.java
@@ -145,10 +145,23 @@ public class Main {
         }
     }
 
+    private void superTokensStarterTemplate() {
+        System.out.println(Logging.ANSI_GREEN + "  ____                       _____     _                  ");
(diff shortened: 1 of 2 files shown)


rishabhpoddar:
@nikhilbhatia08 , unfortunately we cannot accept this PR as our infra (and likely other people's infra) depends on the logs being printed out. This PR modifies the logs raw strings which may cause unforseen issues, for which we won't have the bandwidth to fix.

Thanks for the attempt though, we appreciate it.

nikhilbhatia08:
@rishabhpoddar Thank you for your feedback.

Issue #582, opened 2023-03-16
Make SuperTokens an OAuth and Open ID provider


rishabhpoddar:
Opened this in favour of: https://github.com/supertokens/for-zenhub/issues/108

We want to support:
- client credential flow
- authorization code grant flow
- access and refresh tokens
- open id connect
- refresh token grant flow
- We also want to support https://github.com/authts/oidc-client-ts

Google doc discussion here: 
- https://docs.google.com/document/d/1NPERAitcIIyM9NxBgmKybxjF1PTroze199pNAY1SPZw/edit#heading=h.k7n4z0un6g0l
- https://docs.google.com/document/d/1KUSQcAI96UPRoqATuq8ofFtbC9kqIVFxuE0nh4xVovo/edit

## Test case TODOs
- [ ] Test case 1..

## TODO
- [ ] Flow diagrams
- [ ] CDI repo changes
- [ ] add an api in the core that takes refresh token + client id to return info of the tokens table that is associated with the refresh token
- [ ] add an api in the core that takes auth code + client id to return the info of the tokens table that is associated with the auth code.
- [ ] add another column in tokens table like “client_secret_required_for_refresh” which will be true if NOT using PKCE.
- [ ] the scopes set in the createTokens endpoint should be a subset of the ones set in the authorization code if using the auth code flow to create tokens.
- [ ] What happens if you have taken a lock on the tokens table based on refresh token hash and client id, and then in parallel, the client id is removed from the db:
   - Will the lock fail? But what if the lock is taken first?
   - Will the removing of client id wait for all the locks to be released.
   - Will both go ahead, but the commit of of lock query fail and it will retry? 
- [ ] Add a new field called `grantType` to `OAuth2TokenInfo`. The grantType field should be an enum that includes the following values: AUTH_CODE, REFRESH_TOKEN, AUTH_CODE_PKCE, and CLIENT_CREDENTIALS.
- [ ] Add `query_string` field of type TEXT to  auth_code and access_token tables
- [ ] Update the cdi spec and flow diagram of create auth_code, create access_token, verify access_token  to include `query_string` param in request as well as response.
- [ ] Write down the list of all response status codes across all APIs
- [ ] `buildAccessToken` and `buildIdToken` recipe functions should also return a  useDynamicSigningKey arg (which will be true by default).
[...]

PR #1010, opened 2024-07-08
feat: OAuth provider support

rishabhpoddar:
## Summary of change

Adds oauth provider feature

## Related issues

- https://github.com/supertokens/supertokens-core/issues/582

## Test Plan

TODO

## Documentation changes

TODO

    - In `build.gradle`

- [x]  license key should have `"oauth"` in it.
- [ ] https://supertokens.slack.com/archives/D02B887A2MQ/p1719875078076949
- [x] check license key fro oauth in the apis
- [ ] in authorization get, check that state is at least 8 chars long
- [ ] Enforce in config check that the oauth urls dont have any path, and normalise them.
- [x] Oauth stuff is to be per app config, and not config yaml config. Cause we should be able to modify app config on the fly without restarting the core for a user. But this config should not be visible in the dashboard.
- [ ] Add tests for new function to make PATCH API calls (new HttpRequest function is added)


rishabhpoddar:
# APIs that need to be added to the core

- authorization GET endpoint: https://www.ory.sh/docs/hydra/reference/api#tag/oAuth2/operation/oAuth2Authorize
   - This one gets and sets csrf cookies.
   - we need to take (and forward) tenantId information in the input of this on the backend (
   - the core should validate that the client id exists for this customer
   - Also does redirection through 302s (Location header)
   - Change to POST
      - Take cookie values as body
      - take the query params as query params as a single string in the body
   - return a JSON body with:
      - redirectTo:
         - url created by hydra replace domain by something like [apidomain]
      - cookies : an array with the information we need to set for cookies
      - ideally we want to remove the ory_ prefix if present in tokens
- tokenPOST endpoint: https://www.ory.sh/docs/hydra/reference/api#tag/oAuth2/operation/oauth2TokenExchange
   - This one takes a form post body
      - client_id
      - redirect_uri
      - code
      - grant_type
      - client_secret
      - code_verifier
   - Ideally it should take and return a JSON body
   - The keys can remain in snake case since the backend API has to provide a standard interface
- create/get/delete client endpoints
   - hydra takes & returns snake_case keys, which we transform
   - there are tons of client props, we may not need to support them all but I see no strong reason to skip them either
   - clientId should have a pre-defined shape so we can validate it in the OAuth2Client recipe init
   - https://www.ory.sh/docs/hydra/reference/api#tag/oAuth2/operation/createOAuth2Client
   - https://www.ory.sh/docs/hydra/reference/api#tag/oAuth2/operation/getOAuth2Client
   - https://www.ory.sh/docs/hydra/reference/api#tag/oAuth2/operation/deleteOAuth2Client
- update clients https://www.ory.sh/docs/hydra/reference/api#tag/oAuth2/operation/patchOAuth2Client
   - their patch endpoint has an interesting syntax (taking an array typed as { from: string; op: "replace"; path: string; value: any } )
   - we work around it (simplify it) in the BE: https://github.com/supertokens/supertokens-node/blob/ff2135b542a37fec07af44e9d72467e1304cc147/lib/ts/recipe/oauth2/recipeImplementation.ts#L355, this could be moved to the core, but not strictly necessary
- list clients
   - the next token is defined like https://github.com/supertokens/supertokens-node/blob/ff2135b542a37fec07af44e9d72467e1304cc147/lib/ts/recipe/oauth2/recipeImplementation.ts#L310
   - ideally we'd want it as a part of the response body
- get/accept/reject loginRequest
   - hydra takes & returns snake_case keys, which we transform
   - we can think about how we can build the redirection urls on the BE
- get/accept/reject consentRequest
   - hydra takes & returns snake_case keys, which we transform
   - we can think about how we can build the redirection urls on the BE
- ~get/accept/reject logoutRequest~
   - ~hydra takes & returns snake_case keys, which we transform~
   - ~we can think about how we can build the redirection urls on the BE~
- revoke all oauth2 tokens for user: https://www.ory.sh/docs/hydra/reference/api#tag/oAuth2/operation/oauth2TokenExchange
- revoke specific tokens: https://www.ory.sh/docs/hydra/reference/api#tag/oAuth2/operation/revokeOAuth2Token
- token introspection https://www.ory.sh/docs/hydra/reference/api#tag/oAuth2/operation/oauth2TokenExchange

PR #1018, opened 2024-07-29
Oauth - WIP

tamassoltesz:
## Summary of change

## Related issues

- Link to issue1 here
- Link to issue1 here

## Test Plan

## Documentation changes

    - In `build.gradle`

- [ ] Item1
- [ ] Item2


src/main/java/io/supertokens/config/CoreConfig.java
```diff
@@ -275,15 +275,16 @@ public class CoreConfig {
...
     @HideFromDashboard
     @ConfigDescription(
             "If specified, the core uses this URL to connect to the OAuth provider public service.")
     private String oauth_provider_public_service_url = null;
 
-    @ConfigYamlOnly
+    @NotConflictingInApp
     @JsonProperty
+    @HideFromDashboard
     @ConfigDescription(
```
rishabhpoddar, line 288: Need to modify the normalise function in this file to do lower case for the urls?

src/main/java/io/supertokens/inmemorydb/Start.java
```diff
@@ -102,7 +103,7 @@ public class Start
         implements SessionSQLStorage, EmailPasswordSQLStorage, EmailVerificationSQLStorage, ThirdPartySQLStorage,
         JWTRecipeSQLStorage, PasswordlessSQLStorage, UserMetadataSQLStorage, UserRolesSQLStorage, UserIdMappingStorage,
         UserIdMappingSQLStorage, MultitenancyStorage, MultitenancySQLStorage, TOTPSQLStorage, ActiveUsersStorage,
-        ActiveUsersSQLStorage, DashboardSQLStorage, AuthRecipeSQLStorage {
+        ActiveUsersSQLStorage, DashboardSQLStorage, AuthRecipeSQLStorage, OAuthStorage {
```
rishabhpoddar, line 106: Need to create an interface for OAuthSQLStorage which extends OAuthStorage, and then have this class implement OAuthSQLStorage

src/main/java/io/supertokens/oauth/OAuth.java
```diff
@@ -16,40 +16,88 @@
...
+                String allCookies = responseHeaders.get(COOKIES_HEADER_NAME);
+                cookies = Arrays.asList(allCookies.split("; "));
+            }
         }
 
-        // TODO: parse url resposne and send appropriate reply from this API.
+        return new OAuthAuthResponse(redirectTo, cookies);
+    }
+
+    private static Map<String, String> constructHydraRequestParams(String clientId,
```
rishabhpoddar, line 93: There are going to be other requests to Hydra in this file too. So maybe a better function name would be constructHydraRequestParamsForAuthorizationGETAPICall

src/main/java/io/supertokens/oauth/OAuth.java
```diff
@@ -16,40 +16,88 @@
...
-            redirectTo = Config.getBaseConfig(main).getOAuthProviderPublicServiceUrl() +
+            redirectTo =  publicOAuthProviderServiceUrl +
                     "/oauth2/fallbacks/error?error=invalid_client&error_description=Client+authentication+failed+%28e" +
                     ".g.%2C+unknown+client%2C+no+client+authentication+included%2C+or+unsupported+authentication" +
                     "+method%29.+The+requested+OAuth+2.0+Client+does+not+exist.";
         } else {
             // we query hydra
+            Map<String, String> queryParamsForHydra = constructHydraRequestParams(clientId, redirectURI, responseType, scope, state);
+            Map<String, String> responseHeaders = new HashMap<>();
+            HttpRequest.sendGETRequestWithResponseHeaders(main, null, Config.getBaseConfig(main).getOAuthProviderPublicServiceUrl(), queryParamsForHydra, 20, 400, null, responseHeaders); // TODO is there some kind of config for the timeouts?
```
rishabhpoddar, line 69: Answering your TODO: No need to have a config for the timeout. We can just set it to a high number.

src/main/java/io/supertokens/oauth/OAuth.java
```diff
@@ -16,40 +16,88 @@
...
+
```
rishabhpoddar, line 70: we need to first check if the response status is 302? Maybe Hydra returns different response status if something else is wrong?

src/main/java/io/supertokens/oauth/OAuth.java
```diff
@@ -16,40 +16,88 @@
...
-        //     - localhost:3000, then we redirect to apiDomain
-        //     - public url for hydra, then we throw a 400 error with the right json
-        //     - else we redirect back to the client
 
         OAuthStorage oauthStorage = StorageUtils.getOAuthStorage(storage);
 
         String redirectTo = null;
+        List<String> cookies = null;
+
+        String publicOAuthProviderServiceUrl = Config.getBaseConfig(main).getOAuthProviderPublicServiceUrl();
```
rishabhpoddar, line 58: if this is null (not configured by the user), we need to throw an exception like BadInputException (which will eventually yield a 400 error from the core).

rishabhpoddar: Also, using getBaseConfig(main) is not the right thing todo, since that does not load the config for the current app that's querying this API. Instead, you want to use `Config.getConfig(appIdentifier.toTenantIdentifier())` (or whatever the function name is called in the AppIdentifier class)

src/main/java/io/supertokens/oauth/OAuth.java
```diff
@@ -16,40 +16,88 @@
...
             // we query hydra
+            Map<String, String> queryParamsForHydra = constructHydraRequestParams(clientId, redirectURI, responseType, scope, state);
+            Map<String, String> responseHeaders = new HashMap<>();
+            HttpRequest.sendGETRequestWithResponseHeaders(main, null, Config.getBaseConfig(main).getOAuthProviderPublicServiceUrl(), queryParamsForHydra, 20, 400, null, responseHeaders); // TODO is there some kind of config for the timeouts?
+
+            if(!responseHeaders.isEmpty() && responseHeaders.containsKey(LOCATION_HEADER_NAME)) {
+                String locationHeaderValue = responseHeaders.get(LOCATION_HEADER_NAME);
+
+                if (locationHeaderValue.equals(publicOAuthProviderServiceUrl)){
+                    throw new OAuthException();
```
rishabhpoddar, line 75: throw a custom type of exception which is descriptive for this case. It should map to a 400 error from the core with the right JSON content (based on the query params in the location header). OAuthException is not a very descriptive name for this case.

src/main/java/io/supertokens/oauth/OAuth.java
```diff
@@ -16,40 +16,88 @@
...
         } else {
             // we query hydra
+            Map<String, String> queryParamsForHydra = constructHydraRequestParams(clientId, redirectURI, responseType, scope, state);
+            Map<String, String> responseHeaders = new HashMap<>();
+            HttpRequest.sendGETRequestWithResponseHeaders(main, null, Config.getBaseConfig(main).getOAuthProviderPublicServiceUrl(), queryParamsForHydra, 20, 400, null, responseHeaders); // TODO is there some kind of config for the timeouts?
+
+            if(!responseHeaders.isEmpty() && responseHeaders.containsKey(LOCATION_HEADER_NAME)) {
+                String locationHeaderValue = responseHeaders.get(LOCATION_HEADER_NAME);
+
+                if (locationHeaderValue.equals(publicOAuthProviderServiceUrl)){
```
rishabhpoddar, line 74: We have to be careful here:
- There could be a trailing `/` configured by the user in the config for the url, and hydra may not return that in the locationHeaderValue.
- `locationHeaderValue` returned from hydra is going to have a path and query params and everything int he url, whereas `publicOAuthProviderServiceUrl` configured by the user will not have any path. So the `.equals` will never be true.
- If the user configures `publicOAuthProviderServiceUrl` as a service name (for example if they are using docker compose), then the url returned from hydra response won't match. Maybe we need a different core config like `publicOAuthProviderServiceUrlConfiguredInHydra` - which by default will be the same as the `publicOAuthProviderServiceUrl` config. And we can also assert in validateAndNormalise functino (on coreConfig.java) that `publicOAuthProviderServiceUrlConfiguredInHydra` always resolves to a URL.

src/main/java/io/supertokens/oauth/OAuth.java
```diff
@@ -16,40 +16,88 @@
...
+                    throw new OAuthException();
+                }
+
+                if (locationHeaderValue.equals("localhost:3000")) {
+                    redirectTo = Multitenancy.getAPIDomain(storage, appIdentifier);
```
rishabhpoddar, line 79: I think there is a miscommunication here. We just want the redirectTo string to have the domain of `{apiDomain}`. Like the literal string `{apiDomain}/...` and not the value of apiDomain.

src/main/java/io/supertokens/webserver/api/oauth/OAuthAPI.java
```diff
@@ -0,0 +1,98 @@
...
+import io.supertokens.webserver.WebserverAPI;
+import jakarta.servlet.ServletException;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
+import org.apache.catalina.Store;
+
+import java.io.IOException;
+import java.io.Serial;
+
+public class OAuthAPI extends WebserverAPI {
```
rishabhpoddar, line 45: A better name for this class please.

src/main/java/io/supertokens/webserver/api/oauth/OAuthAPI.java
```diff
@@ -0,0 +1,98 @@
...
+    @Serial
+    private static final long serialVersionUID = -8734479943734920904L;
+
+    public OAuthAPI(Main main) {
+        super(main, RECIPE_ID.OAUTH.toString());
+    }
+
+    @Override
+    public String getPath() {
+        return "oauth/auth";
```
rishabhpoddar, line 55:
```suggestion
        return "/recipe/oauth/auth";
```

src/main/java/io/supertokens/webserver/api/oauth/OAuthAPI.java
```diff
@@ -0,0 +1,98 @@
...
+        JsonObject input = InputParser.parseJsonObjectOrThrowError(req);
+        String clientId = InputParser.parseStringOrThrowError(input, "clientId", false);
+        String redirectUri = InputParser.parseStringOrThrowError(input, "redirectUri", false);
+        String responseType = InputParser.parseStringOrThrowError(input, "responseType", false);
+        String scope = InputParser.parseStringOrThrowError(input, "scope", false);
+        String state = InputParser.parseStringOrThrowError(input, "state", false);
+
+        try {
+            AppIdentifier appIdentifier = getAppIdentifier(req);
+            Storage storage = getTenantStorage(req);
```
rishabhpoddar, line 71: this needs to change to something like this: https://github.com/supertokens/supertokens-core/blob/master/src/main/java/io/supertokens/webserver/api/userroles/CreateRoleAPI.java#L86 (line 86-87)

src/main/java/io/supertokens/webserver/api/oauth/OAuthAPI.java
```diff
@@ -0,0 +1,98 @@
...
+
+            OAuthAuthResponse authResponse = OAuth.getAuthorizationUrl(super.main, appIdentifier, storage,
+                    clientId, redirectUri, responseType, scope, state);
+            JsonObject response = new JsonObject();
+            response.addProperty("redirectTo", authResponse.redirectTo);
+
+            if (authResponse.cookies != null) {
+                Gson gson = new Gson();
```
rishabhpoddar, line 79: We use JSONObject instead of Gson. https://github.com/supertokens/supertokens-core/blob/master/src/main/java/io/supertokens/webserver/api/userroles/CreateRoleAPI.java#L91. Please follow existing patterns as much as possible

src/main/java/io/supertokens/webserver/api/oauth/OAuthAPI.java
```diff
@@ -0,0 +1,98 @@
...
+                String cookiesAsJson = gson.toJson(authResponse.cookies);
+                response.addProperty("cookies", cookiesAsJson);
+            }
+
+            super.sendJsonResponse(200, response, resp);
+
+        } catch (OAuthException e) {
+
+            JsonObject errorResponse = new JsonObject();
+            //TODO what is a good content here?
```
rishabhpoddar, line 89: We need to send a 400 response to the user, with the right JSON (based on the redirectTo link). The redirectTo link will have query params that describe the issue, so that needs to be converted to a JSON, so you need to figure out what those possibilities are - somehow.

src/test/java/io/supertokens/test/oauth/api/OAuthAPITest.java
```diff
@@ -0,0 +1,102 @@
...
+
+        this.process = TestingProcessManager.start(args);
+        FeatureFlagTestContent.getInstance(process.getProcess())
+                .setKeyValue(FeatureFlagTestContent.ENABLED_FEATURES, new EE_FEATURES[]{EE_FEATURES.OAUTH});
+        process.startProcess();
+        assertNotNull(process.checkOrWaitForEvent(ProcessState.PROCESS_STATE.STARTED));
+
+        if (StorageLayer.getStorage(process.getProcess()).getType() != STORAGE_TYPE.SQL) { //TODO check if this is true here also
+            return;
+        }
```
rishabhpoddar, line 66: This is true, but adding it here won't skip any of the tests anyway?

src/test/java/io/supertokens/test/oauth/api/OAuthAPITest.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @After
+    public void afterEach() throws InterruptedException {
+        process.kill();
+        assertNotNull(process.checkOrWaitForEvent(ProcessState.PROCESS_STATE.STOPPED));
+    }
+
+    @Test
+    public void test()
```
rishabhpoddar, line 76: Please name test better, also, please follow existing pattern of other tests - we start and stop the core in the test itself.

src/test/java/io/supertokens/test/oauth/api/OAuthAPITest.java
```diff
@@ -0,0 +1,102 @@
...
+        String responseType = "code";
+        String scope = "profile";
+        String state = "%EF%BF%BD%EF%BF%BD%EF%BF%BD%EF%BF%BD%EF%BF%BD%EF%BF%BDv%EF%BF%BD%EF%BF%BD%EF%BF%BD%EF%BF%BD%EF%BF%BD";
+
+        OAuthStorage oAuthStorage = (OAuthStorage) StorageLayer.getStorage(process.getProcess());
+
+        OAuthAuthResponse response = OAuth.getAuthorizationUrl(process.getProcess(), null, oAuthStorage,  clientId, redirectUri, responseType, scope, state);
+
+        System.out.println(response);
+        System.out.println(response.redirectTo);
```
rishabhpoddar, line 91: no system.out.println please.

src/test/java/io/supertokens/test/oauth/api/OAuthAPITest.java
```diff
@@ -0,0 +1,102 @@
...
+
+        for(String cooke : response.cookies){
+            System.out.println(cooke);
+        }
+
+        assertNotNull(response);
+        assertNotNull(response.redirectTo);
+        assertNotNull(response.cookies);
```
rishabhpoddar, line 99: need to assert actual values and not just that they are not null

rishabhpoddar (changes requested):

src/main/java/io/supertokens/config/CoreConfig.java
tamassoltesz: According to google and w3c, urls should be case sensitive (https://www.w3.org/TR/WD-html40-970708/htmlweb.html). For this reason I would rather not lowecase them

src/main/java/io/supertokens/inmemorydb/Start.java
tamassoltesz: Yeah, that's the plan for today

src/main/java/io/supertokens/oauth/OAuth.java
tamassoltesz: If that's null, the getOAuthProviderPublicServiceUrl will throw an InvalidConfigException

src/main/java/io/supertokens/oauth/OAuth.java
tamassoltesz: I feel like I still miss something. Where does this `localhost:3000` come from? This - for now - looks like some magic address. 
I suspect from the example that the `redirectTo` should not be just `{apiDomain}` here. What's the continuation? Is the task here is to change the "localhost:3000" to "{apiDomain}" in the locationHeaderValue?

src/test/java/io/supertokens/test/oauth/api/OAuthAPITest.java
tamassoltesz: That was just for myself, for quick testing. Removed

src/test/java/io/supertokens/test/oauth/api/OAuthAPITest.java
tamassoltesz: Will do when actually working on the tests, not just using them as a trigger point

src/main/java/io/supertokens/oauth/OAuth.java
tamassoltesz: renamed

src/main/java/io/supertokens/oauth/OAuth.java
tamassoltesz: > Also, using getBaseConfig(main) is not the right thing todo, since that does not load the config for the current app that's querying this API. Instead, you want to use `Config.getConfig(appIdentifier.toTenantIdentifier())` (or whatever the function name is called in the AppIdentifier class)

Done

src/main/java/io/supertokens/webserver/api/oauth/OAuthAPI.java
tamassoltesz: Done

src/main/java/io/supertokens/webserver/api/oauth/OAuthAPI.java
tamassoltesz: Done

src/main/java/io/supertokens/webserver/api/oauth/OAuthAPI.java
tamassoltesz: Done. For now it returns the `error` and `error_description` values.

src/main/java/io/supertokens/httpRequest/HttpRequest.java
```diff
@@ -152,12 +152,12 @@ public static <T> T sendGETRequestWithResponseHeaders(Main main, String requestI
             if (version != null) {
                 con.setRequestProperty("api-version", version + "");
             }
-
+            con.setInstanceFollowRedirects(false);
```
rishabhpoddar, line 155: This should be an input param, cause we use this same function in other places too, and we wouldn't want to change the behaviour of those functions.

src/main/java/io/supertokens/httpRequest/HttpRequest.java
```diff
@@ -152,12 +152,12 @@ public static <T> T sendGETRequestWithResponseHeaders(Main main, String requestI
...
             int responseCode = con.getResponseCode();
 
             con.getHeaderFields().forEach((key, value) -> {
                 if (key != null) {
-                    responseHeaders.put(key, value.get(0));
+                    responseHeaders.put(key, value.get(0)); // TODO why the first element only? What happens with Set-Cookie headers? (Those are repeated if there are multiple cookies)
```
rishabhpoddar, line 160: Answering the TODO: This is a bug in the existing code. Fixing this would also mean that we can't use `key` in the map, cause value.get(1) would override the previous entry in the key. So it would perhaps need to change to map(string, string[])

src/main/java/io/supertokens/oauth/OAuth.java
```diff
@@ -18,38 +18,100 @@
...
-                    "/oauth2/fallbacks/error?error=invalid_client&error_description=Client+authentication+failed+%28e" +
-                    ".g.%2C+unknown+client%2C+no+client+authentication+included%2C+or+unsupported+authentication" +
-                    "+method%29.+The+requested+OAuth+2.0+Client+does+not+exist.";
+            throw new OAuthAuthException("invalid_client", "Client authentication failed (e.g., unknown client, no client authentication included, or unsupported authentication method). The requested OAuth 2.0 Client does not exist.");
         } else {
             // we query hydra
+            Map<String, String> queryParamsForHydra = constructHydraRequestParamsForAuthorizationGETAPICall(clientId, redirectURI, responseType, scope, state);
+            Map<String, String> responseHeaders = new HashMap<>();
+
+            //TODO maybe check response status code? Have to modify sendGetRequest.. for that
```
rishabhpoddar, line 65: In general, if the response code is < 400, the function doesn't throw, else it throws. I think based on the API spec, it would return a 302, or then >= 400 (but this needs to be tested properly). If it's true, we dont need to change the function.

src/main/java/io/supertokens/oauth/OAuth.java
```diff
@@ -18,38 +18,100 @@
...
+                String locationHeaderValue = responseHeaders.get(LOCATION_HEADER_NAME);
+
+                if (locationHeaderValue.contains(publicOAuthProviderServiceUrl)){
+                    String error = getValueOfQueryParam(locationHeaderValue, ERROR_LITERAL);
+                    String errorDescription = getValueOfQueryParam(locationHeaderValue, ERROR_DESCRIPTION_LITERAL);
+                    throw new OAuthAuthException(error, errorDescription);
+                }
+
+                if (locationHeaderValue.contains("localhost:3000") || locationHeaderValue.contains("127.0.0.1:3000")) {
+                    redirectTo = locationHeaderValue.replace("localhost:3000", "{apiDomain}").replace("127.0.0.1:3000", "{apiDomain}");
```
rishabhpoddar, line 78: Need to confirm this with @porcellus - should the resulting string be `http://{apiDomain}/...` or be `{apiDomain}/...` - what does the backend SDK expect?

src/main/java/io/supertokens/oauth/OAuth.java
```diff
@@ -18,38 +18,100 @@
...
+            queryParam = queryParam + "=";
+        }
+        int startIndex = url.indexOf(queryParam) + queryParam.length(); // start after the '=' sign
+        int endIndex = url.indexOf("&", startIndex);
+        if (endIndex == -1){
+            endIndex = url.length();
+        }
+        valueOfQueryParam = url.substring(startIndex, endIndex); // substring the url from the '=' to the next '&' or to the end of the url if there are no more &s
+        return URLDecoder.decode(valueOfQueryParam, StandardCharsets.UTF_8);
     }
```
rishabhpoddar, line 116: Maybe instead of manually parsing it, we should use some lib that auto parses a url and returns a query param? This is fine too though, as long as it's extensively tested.

src/main/java/io/supertokens/oauth/OAuth.java
rishabhpoddar: I see that you used `.contains` now instead of `.equals`, which is better. However, it could cause issues if the `locationHeaderValue` returned from hydra is something like `https://example.com?...` and the value configured for `publicOAuthProviderServiceUrl` is `https://example.com/`. The difference here is the trailing `/`. 

It may be a better idea to normalise URLS first before checking comparing them. One way could be to parse them through some lib that returns a URL object, and then fetch the domain from it and then compare that string with `.equals`. 

URL comparison is tricky, and we need to be careful.

-------

Also, you didn't reply to the docker server point I bought up above - what do you think?

src/main/java/io/supertokens/oauth/OAuth.java
rishabhpoddar: What you have implemented is correct. `localhost:3000` comes from ory (some config in ory is set to that by default). I don't quite recall what it implies. Maybe ask @porcellus. Let me know too :)

src/main/java/io/supertokens/oauth/exceptions/OAuthAuthException.java
```diff
@@ -0,0 +1,30 @@
...
+package io.supertokens.oauth.exceptions;
+
+public class OAuthAuthException extends Exception{
+    private static final long serialVersionUID = 1836718299845759897L;
+
+    public final String error;
+    public final String errorDescription;
+
+    public OAuthAuthException(String error, String errorDescription) {
+        this.error = error;
```
rishabhpoddar, line 26: call the super class to, so that we see a full stack and stuff as well when print out the exception?

src/main/java/io/supertokens/webserver/api/oauth/OAuthAuthAPI.java
```diff
@@ -0,0 +1,97 @@
...
+    @Serial
+    private static final long serialVersionUID = -8734479943734920904L;
+
+    public OAuthAuthAPI(Main main) {
+        super(main, RECIPE_ID.OAUTH.toString());
+    }
+
+    @Override
+    public String getPath() {
+        return "recipe/oauth/auth";
```
rishabhpoddar, line 51:
```suggestion
        return "/recipe/oauth/auth";
```

rishabhpoddar: Idk if it makes a difference, but other apis have `/` in the front too

src/main/java/io/supertokens/webserver/api/oauth/OAuthAuthAPI.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws IOException, ServletException {
+        // TODO Work in progress!
```
rishabhpoddar, line 56: Remove this todo?

src/main/java/io/supertokens/webserver/api/oauth/OAuthAPI.java
rishabhpoddar: Please confirm with @porcellus if this is OK output.

src/test/java/io/supertokens/test/oauth/api/OAuthAPITest.java
rishabhpoddar: Please follow existing patterns of other tests where each test starts and stops the core in the test itself.

src/main/java/io/supertokens/inmemorydb/Start.java
```diff
@@ -3007,4 +3008,23 @@ public int countUsersThatHaveMoreThanOneLoginMethodOrTOTPEnabledAndActiveSince(A
...
+            throws StorageQueryException {
+        try {
+            return OAuthQueries.isClientIdForAppId(this, clientId, appIdentifier);
+        } catch (SQLException e) {
+            throw new StorageQueryException(e);
+        }
+    }
+
+    @Override
+    public void addClientForApp(AppIdentifier appIdentifier, String clientId) throws StorageQueryException {
```
rishabhpoddar, line 3023:
```suggestion
    public void addClientForApp(AppIdentifier appIdentifier, String clientId) throws StorageQueryException, ClientAlreadyExistsException {
```

rishabhpoddar: Since this is an insert query, it can fail if the client already exists, in this case, we need to communicate that to the core logic, which can be communicated via this exception. If you see other queries, the way we detect this is by parsing the error string thrown from the db query based on the primary key

src/main/java/io/supertokens/inmemorydb/queries/OAuthQueries.java
```diff
@@ -0,0 +1,67 @@
...
+import static io.supertokens.inmemorydb.QueryExecutorTemplate.execute;
+import static io.supertokens.inmemorydb.QueryExecutorTemplate.update;
+
+public class OAuthQueries {
+
+    public static String getQueryToCreateOAuthClientTable(Start start) {
+        String oAuth2ClientTable = Config.getConfig(start).getOAuthClientTable();
+        // @formatter:off
+        return "CREATE TABLE IF NOT EXISTS " + oAuth2ClientTable + " ("
+                + "app_id VARCHAR(64) DEFAULT 'public',"
```
rishabhpoddar, line 39:
```suggestion
                + "app_id VARCHAR(64),"
```

rishabhpoddar: Better to not have default since if some query in the code forgets to add it, then it may not be caught

rishabhpoddar (changes requested):

src/main/java/io/supertokens/inmemorydb/queries/OAuthQueries.java
tamassoltesz: yeah, that's valid. Removing it.

src/main/java/io/supertokens/oauth/OAuth.java
tamassoltesz: I agree that we have to first test it.

src/main/java/io/supertokens/oauth/OAuth.java
tamassoltesz: yeah, using a full lib for only this seemd a bit overkill, but I will consider it.

src/main/java/io/supertokens/oauth/exceptions/OAuthAuthException.java
tamassoltesz: a call to the super class's default constructor is automatically inserted by the compiler. But If you would like, I can write it out

src/main/java/io/supertokens/webserver/api/oauth/OAuthAuthAPI.java
tamassoltesz: It doesn't make a difference, as in the WebserverAPI and also in the PathRouter it is checked if the path starts with a `/` and appends it if it doesn't. For the sake of uniformity, I will append it.

src/main/java/io/supertokens/webserver/api/oauth/OAuthAuthAPI.java
tamassoltesz: yeah, seems like

src/main/java/io/supertokens/httpRequest/HttpRequest.java
tamassoltesz: yeah, that's true. For now, it is only used in two tests, for them it makes no difference as far as I can see. On the other hand, this is not the default behaviour of the `HttpURLConnection` so in the future it could be a surprise. I change it to a param.

src/main/java/io/supertokens/inmemorydb/Start.java
tamassoltesz: Thanks! Done

src/test/java/io/supertokens/test/oauth/api/OAuthAPITest.java
tamassoltesz: Yeah, this was based on an other test. It seems like I didn't find the right one to use as an example

src/main/java/io/supertokens/webserver/api/oauth/OAuthAPI.java
tamassoltesz: sure, I will

src/main/java/io/supertokens/oauth/OAuth.java
tamassoltesz: I have to dig deeper into hydra, to answer those. Yesterday I didn't get to it. Now I feel like I am missing some important details. I will do it today.

src/main/java/io/supertokens/oauth/OAuth.java
tamassoltesz: without http

src/main/java/io/supertokens/webserver/api/oauth/OAuthAPI.java
tamassoltesz: Confirmed, it's a good content for errors.

src/main/java/io/supertokens/oauth/OAuth.java
tamassoltesz: This localhost/127.0.0.1 comes from a hydra setting. This is the start of the url for the consent/login/etc hydra endpoints, whis is where you would get the redirect after the auth.
We want to change to apiDomain, so hydra and core doesn't have to worry about where to redirect, the sdks will take care of that. 
I didn't change mine, so my hydra instance sends me to 127.0.0.1, @porcellus set his to localhost. This can differ from setup-to-setup, so I propose making it a config in core.

src/main/java/io/supertokens/oauth/OAuth.java
tamassoltesz: So, regarding the docker service address: the url used by hydra here is the default fallback for now (to `{hydra_self}/oauth2/fallbacks/error`). This indeed can be different from what was set in the core config as publicOAuthProviderServiceUrl, although this could be resolved in the hydra settings too. On the other hand, I would suggest to handle this on the core side, as we have greater control over that.

src/main/java/io/supertokens/oauth/OAuth.java
```diff
@@ -18,38 +18,100 @@
...
+                    String errorDescription = getValueOfQueryParam(locationHeaderValue, ERROR_DESCRIPTION_LITERAL);
+                    throw new OAuthAuthException(error, errorDescription);
+                }
+
+                if(Utils.containsUrl(locationHeaderValue, hydraBaseUrlForConsentAndLogin, true)){
+                    redirectTo = locationHeaderValue.replace(hydraBaseUrlForConsentAndLogin, "{apiDomain}");
+                } else {
+                    redirectTo = locationHeaderValue;
+                }
+            }
```
rishabhpoddar, line 84: in the else case here, what should we do? Probably throw a RuntimeError? Since it's not expected.. and we should not silently return a `null` redirectTo?

src/main/java/io/supertokens/oauth/OAuth.java
rishabhpoddar: Keeping this comment unresolved cause we have to test it first. If you have already done so, please resolve this comment.

rishabhpoddar (changes requested):

src/main/java/io/supertokens/config/CoreConfig.java
```diff
@@ -833,6 +868,43 @@ void normalizeAndValidate(Main main, boolean includeConfigFilePath) throws Inval
...
+            try {
+                URL url = new URL(oauth_provider_public_service_url);
+            } catch (MalformedURLException malformedURLException){
+                throw new InvalidConfigException("oauth_provider_public_service_url is not a valid URL");
+            }
+        }
+
+        if(oauth_provider_admin_service_url != null) {
+            try {
+                URL url = new URL(oauth_provider_admin_service_url);
```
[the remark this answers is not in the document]
tamassoltesz: No, it won't throw an error for that.

src/main/java/io/supertokens/config/CoreConfig.java
```diff
@@ -833,6 +868,43 @@ void normalizeAndValidate(Main main, boolean includeConfigFilePath) throws Inval
...
+            try {
+                URL url = new URL(oauth_provider_consent_login_base_url);
+            } catch (MalformedURLException malformedURLException){
+                throw new InvalidConfigException("oauth_provider_consent_login_base_url is not a valid URL");
+            }
+        }
+
+
+        if(oauth_provider_url_configured_in_hydra == null) {
+            oauth_provider_url_configured_in_hydra = oauth_provider_public_service_url;
```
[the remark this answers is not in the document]
tamassoltesz: I set it there, but that's the point where it is not needed. When instantiating the `CoreConfig` object, the `ConfigMapper` (in `mapConfig`), creates a new instance. At that point both of them are null. After that, the `configMapper` goes through all of the set configs in the config file, setting the `oauth_provider_public_service_url`, but the reference of the `oath_provider_url_configured_in_hydra` doesn't change from the `null` value.

src/main/java/io/supertokens/config/CoreConfig.java
```diff
@@ -833,6 +868,43 @@ void normalizeAndValidate(Main main, boolean includeConfigFilePath) throws Inval
...
+        } else {
+            try {
+                URL url = new URL(oauth_provider_url_configured_in_hydra);
+            } catch (MalformedURLException malformedURLException){
+                throw new InvalidConfigException("oauth_provider_url_configured_in_hydra is not a valid URL");
+            }
+        }
+
```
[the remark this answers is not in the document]
tamassoltesz: except for those which has default values? (So if the public url is set, the admin and the consent have to be set too. The url in hydra defaults to the public, so I wouldn't enforce to set it explicitly)

src/main/java/io/supertokens/inmemorydb/Start.java
```diff
@@ -3007,4 +3009,32 @@ public int countUsersThatHaveMoreThanOneLoginMethodOrTOTPEnabledAndActiveSince(A
...
+        try {
+            return OAuthQueries.isClientIdForAppId(this, clientId, appIdentifier);
+        } catch (SQLException e) {
+            throw new StorageQueryException(e);
+        }
+    }
+
+    @Override
+    public void addClientForApp(AppIdentifier appIdentifier, String clientId)
+            throws StorageQueryException, ClientAlreadyExistsForAppException {
```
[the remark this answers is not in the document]
tamassoltesz: Sure!

src/main/java/io/supertokens/oauth/OAuth.java
tamassoltesz: yeah, that's sounds reasonable. This can only happen if there is no location header in the response from hydra, so that's an unexpected case.

src/main/java/io/supertokens/oauth/OAuth.java
tamassoltesz: I couldn't find any cases where it went above 400. (Generally, 302 or 303). Of course, this doesn't prove that there are _no cases_ where it would return a 400 or above, but I couldn't find any.

rishabhpoddar: In what case does it return a 400?

src/main/java/io/supertokens/inmemorydb/Start.java
```diff
@@ -3032,9 +3032,27 @@ public void addClientForApp(AppIdentifier appIdentifier, String clientId)
...
     }
+
+    @Override
+    public boolean isClientIdAlreadyExists(String clientId) throws StorageQueryException {
+        try {
+            return OAuthQueries.isClientIdAlreadyExists(this, clientId);
+        } catch (SQLException e) {
+            throw new StorageQueryException(e);
+        }
+    }
```
rishabhpoddar, line 3048: everything in the core is per app (if not per tenant) - so why do we need this function given that we already have doesClientIdExistForThisApp?

src/main/java/io/supertokens/inmemorydb/Start.java
```diff
@@ -3032,9 +3032,27 @@ public void addClientForApp(AppIdentifier appIdentifier, String clientId)
...
+
+    @Override
+    public void removeAppClientAssociation(AppIdentifier appIdentifier, String clientId) throws StorageQueryException {
```
rishabhpoddar, line 3051: this should communicate back to the caller if the client was associated with the app in the first place? For example, with a boolean

src/main/java/io/supertokens/oauth/exceptions/OAuthClientException.java
```diff
@@ -0,0 +1,29 @@
...
+ *    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ *    License for the specific language governing permissions and limitations
+ *    under the License.
+ */
+
+package io.supertokens.oauth.exceptions;
+
+import java.io.Serial;
+
+public class OAuthClientException extends  OAuthException{
```
rishabhpoddar, line 21: The name of this exception does not really indicate whats going wrong. Does it mean the client already exists, or client does not exist, or something else?

src/main/java/io/supertokens/oauth/exceptions/OAuthClientRegisterException.java
```diff
@@ -0,0 +1,29 @@
...
+ *    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ *    License for the specific language governing permissions and limitations
+ *    under the License.
+ */
+
+package io.supertokens.oauth.exceptions;
+
+import java.io.Serial;
+
+public class OAuthClientRegisterException extends OAuthException{
```
rishabhpoddar, line 21: what during the registration process has gone wrong for this exception to be thrown? The name has to indicate that clearly

src/main/java/io/supertokens/oauth/exceptions/OAuthException.java
```diff
@@ -0,0 +1,33 @@
...
+ *    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ *    License for the specific language governing permissions and limitations
+ *    under the License.
+ */
+
+package io.supertokens.oauth.exceptions;
+
+import java.io.Serial;
+
+public class OAuthException extends Exception{
```
rishabhpoddar, line 21: I dont think you need this class. All OAuth related exceptions can directly inherit from Exception

src/main/java/io/supertokens/utils/Utils.java
```diff
@@ -444,4 +445,21 @@ public static boolean containsUrl(String urlToCheckIfContains, String whatItCont
...
+                .toLowerCase();
+        return toSnakeCase;
+    }
+
+    public static String snakeCaseToCamelCase(String toCamelCase) {
+        toCamelCase = Pattern.compile("_([a-z])")
+                .matcher(toCamelCase)
+                .replaceAll(m -> m.group(1).toUpperCase());
+        return toCamelCase;
+    }
```
rishabhpoddar, line 464: need tests for these

src/main/java/io/supertokens/oauth/OAuth.java
```diff
@@ -90,6 +102,132 @@ public static OAuthAuthResponse getAuthorizationUrl(Main main, AppIdentifier app
...
+            NoSuchAlgorithmException, StorageQueryException {
+
+        OAuthStorage oauthStorage = StorageUtils.getOAuthStorage(storage);
+        String adminOAuthProviderServiceUrl = Config.getConfig(appIdentifier.getAsPublicTenantIdentifier(), main).getOAuthProviderAdminServiceUrl();
+
+        byte[] idBaseBytes = new byte[48];
+
+        while(true){
+            new SecureRandom().nextBytes(idBaseBytes);
+            String clientId = "supertokens_" + Utils.hashSHA256Base64UrlSafe(idBaseBytes);
```
rishabhpoddar, line 117: why do we have the supertokens_ prefix to client ids? I dont mind it, but just curious.

src/main/java/io/supertokens/oauth/OAuth.java
```diff
@@ -90,6 +102,132 @@ public static OAuthAuthResponse getAuthorizationUrl(Main main, AppIdentifier app
...
+            try {
+                if(oauthStorage.isClientIdAlreadyExists(clientId)){
+                    continue; // restart
+                }
```
rishabhpoddar, line 121: this check doesn't really help cause we can have multiple cores (with their own db) connected to one instance of hydra. Also, i assume that if we try and create a new client id with an existing one, hydra itself will throw an error? So we should detect this case based on what hydra responds with

src/main/java/io/supertokens/oauth/OAuth.java
```diff
@@ -90,6 +102,132 @@ public static OAuthAuthResponse getAuthorizationUrl(Main main, AppIdentifier app
...
+
+    public static JsonObject loadOAuthClient(Main main, AppIdentifier appIdentifier, Storage storage, String clientId)
+            throws TenantOrAppNotFoundException, OAuthClientException, InvalidConfigException, StorageQueryException,
+            IOException {
+        OAuthStorage oauthStorage = StorageUtils.getOAuthStorage(storage);
+
+        String adminOAuthProviderServiceUrl = Config.getConfig(appIdentifier.getAsPublicTenantIdentifier(), main).getOAuthProviderAdminServiceUrl();
+
+        if (!oauthStorage.doesClientIdExistForThisApp(appIdentifier, clientId)) {
+            throw new OAuthClientException("Unable to locate the resource", "");
```
rishabhpoddar, line 151:
```suggestion
            throw new OAuthClientIdDoesNotExistForAppException();
```

src/main/java/io/supertokens/oauth/OAuth.java
```diff
@@ -90,6 +102,132 @@ public static OAuthAuthResponse getAuthorizationUrl(Main main, AppIdentifier app
...
+        } else {
+            try {
+                JsonObject hydraResponse = HttpRequest.sendGETRequest(main, "", adminOAuthProviderServiceUrl + HYDRA_CLIENTS_ENDPOINT + "/" + clientId, null, 10000, 10000, null);
+                return  formatResponseForSDK(hydraResponse);
+            } catch (HttpResponseException e) {
+                try {
+                    throw createCustomExceptionFromHttpResponseException(e, OAuthClientException.class);
+                } catch (NoSuchMethodException | InvocationTargetException | InstantiationException |
+                         IllegalAccessException ex) {
+                    throw new RuntimeException("Something went really wrong!");
```
rishabhpoddar, line 161:
```suggestion
                    throw new RuntimeException(ex);
```

src/main/java/io/supertokens/oauth/OAuth.java
```diff
@@ -90,6 +102,132 @@ public static OAuthAuthResponse getAuthorizationUrl(Main main, AppIdentifier app
...
+
+    public static void deleteOAuthClient(Main main, AppIdentifier appIdentifier, Storage storage, String clientId)
+            throws TenantOrAppNotFoundException, OAuthClientException, InvalidConfigException, StorageQueryException,
+            IOException {
+        OAuthStorage oauthStorage = StorageUtils.getOAuthStorage(storage);
+
+        String adminOAuthProviderServiceUrl = Config.getConfig(appIdentifier.getAsPublicTenantIdentifier(), main).getOAuthProviderAdminServiceUrl();
+
+        if (!oauthStorage.doesClientIdExistForThisApp(appIdentifier, clientId)) {
+            throw new OAuthClientException("Unable to locate the resource", "");
```
rishabhpoddar, line 175:
```suggestion
            throw new OAuthClientIdDoesNotExistForAppException();
```

src/main/java/io/supertokens/oauth/OAuth.java
```diff
@@ -90,6 +102,132 @@ public static OAuthAuthResponse getAuthorizationUrl(Main main, AppIdentifier app
...
+        } else {
+            try {
+                HttpRequest.sendJsonDELETERequest(main, "", adminOAuthProviderServiceUrl + HYDRA_CLIENTS_ENDPOINT + "/" + clientId, null, 10000, 10000, null);
+
+                oauthStorage.removeAppClientAssociation(appIdentifier, clientId);
```
rishabhpoddar, line 180: this needs to happen the other way around, cause if it succeeds in hydra, but fails during supertokens db, it will lead to an inconsistent state. But if it's the other way around, then failing in hydra, and succeeding in supertokens core is ok (its the same case as in when creating, it succeeds in creating in hydra, but fails in adding to supertokens)

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws IOException, ServletException {
+        JsonObject input = InputParser.parseJsonObjectOrThrowError(req);
+
+        boolean containsAllRequired = containsAllRequiredFields(input);
+        boolean containsMoreThanAllowed = containsMoreThanAllowed(input);
+        if(!containsAllRequired || containsMoreThanAllowed){
+            throw new ServletException(new WebserverAPI.BadRequestException("Invalid Json Input"));
```
rishabhpoddar, line 71: need to give more details about whats invalid. From a dev's point of view, it will be a very bad experience to get such a vague error message

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
```diff
@@ -0,0 +1,168 @@
...
+            AppIdentifier appIdentifier = getAppIdentifier(req);
+            Storage storage = enforcePublicTenantAndGetPublicTenantStorage(req);
+
+            JsonObject response = OAuth.loadOAuthClient(main, appIdentifier, storage, clientId);
+            sendJsonResponse(200, response, resp);
+
+        }  catch (OAuthClientException e) {
+            JsonObject errorResponse = createJsonFromException(e);
+            sendJsonResponse(400, errorResponse, resp);
+
```
rishabhpoddar, line 106: i dont think sending a 400 is here expected. This condition happens if there was no client id found for this app. So we should return an output like:

```
{status: "CLIENT_DOES_NOT_EXIST_ERROR"}
```

and for the case of when it does exist, we can return:
```
{status: "OK", ...rest of the json fields}
```

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
```diff
@@ -0,0 +1,168 @@
...
+        if(!containsAllRequired || containsMoreThanAllowed){
+            throw new ServletException(new WebserverAPI.BadRequestException("Invalid Json Input"));
+        }
+
+        try {
+            AppIdentifier appIdentifier = getAppIdentifier(req);
+            Storage storage = enforcePublicTenantAndGetPublicTenantStorage(req);
+
+            JsonObject response = OAuth.registerOAuthClient(super.main, appIdentifier, storage, input);
+            sendJsonResponse(200, response, resp);
```
rishabhpoddar, line 79: response needs to be correctly formatted based on how we have apis in the core like:
```
{status: "OK", ...rest of the fields}
```

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
```diff
@@ -0,0 +1,168 @@
...
+
+        } catch (OAuthClientRegisterException registerException) {
+
+            JsonObject errorResponse = createJsonFromException(registerException);
+            sendJsonResponse(400, errorResponse, resp);
```
rishabhpoddar, line 84: response needs to be correctly formatted based on how we have apis in the core like:
```
{status: "CLIENT_ID_ALREADY_EXISTS_ERROR"|
```

With a status of 200

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
rishabhpoddar: Both with a status of 200

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
```diff
@@ -0,0 +1,168 @@
...
+
+            OAuth.deleteOAuthClient(main, appIdentifier, storage, clientId);
+            JsonObject responseBody = new JsonObject();
+            responseBody.addProperty("status", "OK");
+            sendJsonResponse(200, responseBody, resp);
+
+        }  catch (OAuthClientException e) {
+            JsonObject errorResponse = createJsonFromException(e);
+            sendJsonResponse(400, errorResponse, resp);
+
```
rishabhpoddar, line 131: - When can it come in OAuthClientException? The exception name is very vague.
- We should return something like `{status: "OK" | "CLIENT_DOES_NOT_EXIST_ERROR"}` - both with 200 status code

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
```diff
@@ -0,0 +1,168 @@
...
+        } catch (TenantOrAppNotFoundException | InvalidConfigException | BadPermissionException
+                 | StorageQueryException e){
+            throw new ServletException(e);
+        }
+    }
+
```
rishabhpoddar, line 137: API to modify an existing client id?

src/main/java/io/supertokens/oauth/OAuth.java
tamassoltesz: I couldn't find any. I was referencing 400 because the HttpRequest throws exception above 400.

src/main/java/io/supertokens/inmemorydb/Start.java
tamassoltesz: This is used for checking if the client exists at all, not just for the app. @porcellus said that when registering a new client for an app that client is going to have a different clientID. So this is used when generating the clientId to check if that generated clientId is globally unique.

src/main/java/io/supertokens/inmemorydb/Start.java
tamassoltesz: It could, but currently there is no need for it. I would say to do so, when we want to use that for anything.

src/main/java/io/supertokens/oauth/exceptions/OAuthClientException.java
tamassoltesz: That's more like a generic exception to hold the errors returned by hydra. It's called ..ClientException because it gets thrown when using the client APIs. If we want to use specifically named exceptions (instead of the error message in the `error` and `errorDescription` fields to identify the problem) then we have to discover/know all the possible errors hydra can return and map them to our internal exceptions. I don't think that's feasible. What do you think?

src/main/java/io/supertokens/oauth/exceptions/OAuthClientRegisterException.java
tamassoltesz: This is an exception to handle errors returned by hydra. I can't tell at this point what went wrong or what could go wrong. This name indicates that the error happened while we were using the client registration API of hydra, but the concrete error is in the `error` and `errorDescription` fields.

src/main/java/io/supertokens/oauth/exceptions/OAuthException.java
tamassoltesz: This is used to make sure all the subclasses have `error` and `errorDescription` fields. Hydra generally returns errors as some 400 status code and a json body containing `error` and `error_description` fields. I created a method to automatically map our HttpResponseExeption to the subclasses of this OAuthException, but for that I needed to make sure that all the hydra related exceptions have `error` and `errorDescription` fields.

src/main/java/io/supertokens/utils/Utils.java
tamassoltesz: Sure, I agree!

src/main/java/io/supertokens/oauth/OAuth.java
tamassoltesz: It was @porcellus' request to do so.

src/main/java/io/supertokens/oauth/OAuth.java
tamassoltesz: Yeah, I didn't think about the multiple cores to the same hydra scenario. But this at least can guarantee that we don't try to register a new client with already existing clientId of which we know.
About the multiple cores to the same hydra instance, should we silently retry if hydra response is something like "Client already exists?"
(In that case hydra returns `409 Conflict` with a body of `{
    "error": "Unable to insert or update resource because a resource with that value exists already",
    "error_description": ""
}` )

src/main/java/io/supertokens/oauth/OAuth.java
tamassoltesz: Okay. The other way around also leads to an inconsistent state (but at least we treat the client as non existent)

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
tamassoltesz: I didn't get to it yesterday, working on it today :)

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
tamassoltesz: Okay, I will change it to be more specific

src/main/java/io/supertokens/inmemorydb/Start.java
tamassoltesz: Also, if we have one hydra instance, clientIds have to be unique within that hydra instance, so that means that we have to check for globall uniqueness here, not just for app

src/main/java/io/supertokens/oauth/exceptions/OAuthClientException.java
tamassoltesz: Okay, after todays discussion, this was renamed.

src/main/java/io/supertokens/oauth/exceptions/OAuthClientRegisterException.java
tamassoltesz: renamed

src/main/java/io/supertokens/oauth/OAuth.java
tamassoltesz: Do we really need to differentiate the "not found for this app" from the "not found at all"? Currently, I can't see how that would be a help for us, the result in either case would be a "Couldn't find" (Does it really matter if the clientId exists for an other app, but not for the current one?) 
I renamed the `OAuthClientException` to `OAuthClientNotFoundException` so that way the same type of exception can be used to communicate this and the case where the clientId is not found in the oauth provider's database.

src/main/java/io/supertokens/oauth/OAuth.java
tamassoltesz: done

src/main/java/io/supertokens/oauth/OAuth.java
tamassoltesz: done

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
tamassoltesz: done. In the error message I list all the missing required fields. I decided that its not required to check for the `allowed` fields as the core just passes them through to the oauth provider, which ignores the fields that are not needed.

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
tamassoltesz: done

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
tamassoltesz: done

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
tamassoltesz: based on todays discussion, in this case of client already exists, the core should silentlry retry. that's done.

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
tamassoltesz: renamed the exception, changed the status code

src/main/java/io/supertokens/webserver/InputParser.java
```diff
@@ -236,4 +239,19 @@ public static Integer parseIntOrThrowError(JsonObject element, String fieldName,
 
     }
 
+    public static List<String> collectAllMissingRequiredFieldsAndThrowError(JsonObject input, List<String> requiredFields)
```
rishabhpoddar, line 242:
```suggestion
    public static List<String> collectAllMissingRequiredFieldsElseThrowError(JsonObject input, List<String> requiredFields)
```

src/main/java/io/supertokens/webserver/api/oauth/OAuthAuthAPI.java
```diff
@@ -76,16 +78,16 @@ protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws I
...
 
         } catch (OAuthAuthException authException) {
 
             JsonObject errorResponse = new JsonObject();
             errorResponse.addProperty("error", authException.error);
-            errorResponse.addProperty("error_description", authException.errorDescription);
-
-            super.sendJsonResponse(400, errorResponse, resp);
+            errorResponse.addProperty("errorDescription", authException.errorDescription);
+            errorResponse.addProperty("status", "OAUTH2_AUTH_ERROR");
```
rishabhpoddar, line 89: The comment [here](https://github.com/supertokens/supertokens-core/issues/582#issuecomment-2265144074) says `AUTH_ERROR`, however in the code it's `OAUTH2_AUTH_ERROR` - which one is correct? Please modify the comment or the code.

src/main/java/io/supertokens/webserver/api/oauth/OAuthAuthAPI.java
```diff
@@ -46,26 +48,26 @@ public OAuthAuthAPI(Main main) {
...
     @Override
     protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws IOException, ServletException {
+
         JsonObject input = InputParser.parseJsonObjectOrThrowError(req);
-        String clientId = InputParser.parseStringOrThrowError(input, "clientId", false);
-        String redirectUri = InputParser.parseStringOrThrowError(input, "redirectUri", false);
-        String responseType = InputParser.parseStringOrThrowError(input, "responseType", false);
-        String scope = InputParser.parseStringOrThrowError(input, "scope", false);
-        String state = InputParser.parseStringOrThrowError(input, "state", false);
+        InputParser.collectAllMissingRequiredFieldsAndThrowError(input, REQUIRED_FIELDS_FOR_POST);
```
rishabhpoddar, line 63: should we normalise the input as well? for example, making clientId all lower case and trimming it

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
```diff
@@ -41,17 +40,14 @@
...
 
     @Serial
     private static final long serialVersionUID = -4482427281337641246L;
 
-    private static final List<String> ALLOWED_INPUT_FIELDS = Arrays.asList(new String[]{"clientName","scope", "redirectUris", "allowedCorsOrigins", "authorizationCodeGrantAccessTokenLifespan", "authorizationCodeGrantIdTokenLifespan", "authorizationCodeGrantRefreshTokenLifespan",
-        "clientCredentialsGrantAccessTokenLifespan","implicitGrantAccessTokenLifespan","implicitGrantIdTokenLifespan","refreshTokenGrantAccessTokenLifespan","refreshTokenGrantIdTokenLifespan","refreshTokenGrantRefreshTokenLifespan","tokenEndpointAuthMethod","audience",
-        "grantTypes","responseTypes","clientUri","logoUri","policyUri","tosUri","metadata"});
-    private static final List<String> REQUIRED_INPUT_FIELDS = Arrays.asList(new String[]{"clientName", "scope"});
+    private static final List<String> REQUIRED_INPUT_FIELDS_FOR_POST = Arrays.asList(new String[]{"clientName", "scope"});
+    public static final String OAUTH_2_CLIENT_NOT_FOUND = "OAUTH2_CLIENT_NOT_FOUND";
```
rishabhpoddar, line 50:
```suggestion
    public static final String OAUTH_2_CLIENT_NOT_FOUND_ERROR = "OAUTH2_CLIENT_NOT_FOUND_ERROR";
```

rishabhpoddar: We add a postfix of `_ERROR` for non OK status always

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
```diff
@@ -63,25 +59,25 @@ public OAuthClientsAPI(Main main){
...
+            postResponseBody.add("client", response);
+            sendJsonResponse(200, postResponseBody, resp);
 
-        } catch (OAuthClientRegisterException registerException) {
+        } catch (OAuthClientRegisterInvalidInputException registerException) {
 
             JsonObject errorResponse = createJsonFromException(registerException);
-            sendJsonResponse(400, errorResponse, resp);
+            errorResponse.addProperty("status", "INVALID_INPUT");
+            sendJsonResponse(200, errorResponse, resp);
```
rishabhpoddar, line 80: this can happen cause of a programming error, so this should return a 400 to the client. Please also correct the comment in the GH issue

src/main/java/io/supertokens/inmemorydb/Start.java
rishabhpoddar: Yea, but Hydra also does this check and returns a 409 anyway. So i see no reason that we also do this check

src/main/java/io/supertokens/inmemorydb/Start.java
rishabhpoddar: I think it's important to communicate it here cause of Developer experience. If they are deleting a client and by mistake make a typo, then they will get back the same response in case when they have actually deleted a client. So they may be under the wrong impression that the client is deleted (in case they made a typo)

src/main/java/io/supertokens/utils/Utils.java
rishabhpoddar: I dont see any tests added for this, so keeping this open

src/main/java/io/supertokens/oauth/OAuth.java
rishabhpoddar: Yup. So if hydra already tells us about this issue, i don't see why we need to also check for it

src/main/java/io/supertokens/oauth/OAuth.java
```diff
@@ -16,40 +16,242 @@
...
+    }
+
+    //This more or less acts as a pass-through for the sdks, apart from camelCase <-> snake_case key transformation and setting a few default values
+    public static JsonObject registerOAuthClient(Main main, AppIdentifier appIdentifier, Storage storage, JsonObject paramsFromSdk)
+            throws TenantOrAppNotFoundException, InvalidConfigException, IOException,
+            OAuthClientRegisterInvalidInputException,
+            NoSuchAlgorithmException, StorageQueryException {
+
+        OAuthStorage oauthStorage = StorageUtils.getOAuthStorage(storage);
+        String adminOAuthProviderServiceUrl = Config.getConfig(appIdentifier.getAsPublicTenantIdentifier(), main).getOAuthProviderAdminServiceUrl();
```
rishabhpoddar, line 110: usually this will also have an API key associated with it. So that needs to go in the config as well, and used to query the admin endpoint. Please test this out in your local setup.

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
rishabhpoddar: I dont see this in the PR either. Maybe you forgot to push?

src/main/java/io/supertokens/inmemorydb/Start.java
tamassoltesz: I was thinking skipping the network roundtrip when we could detect this in place. But okay, if that's not a concern, I will remove this and the related parts.

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
tamassoltesz: No, I didn't got to it. That will be in a different PR

src/main/java/io/supertokens/webserver/api/oauth/OAuthAuthAPI.java
tamassoltesz: Sure, updated the comment. I think being more specific in the error status (OAUTH2_AUTH_ERROR instead of just (AUTH_ERROR) is a good thing.

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
tamassoltesz: noted, fixed

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
tamassoltesz: corrected. Also, added the `_ERROR` postfix

src/main/java/io/supertokens/oauth/OAuth.java
tamassoltesz: Hm, I'm pretty sure that I didn't set up an API call, and yes, all my tests are using my local hydra instance. I am investigating it further and will update here.

tamassoltesz: This: https://www.ory.sh/docs/hydra/self-hosted/production states that 
```The administrative port shouldn't be exposed to public internet traffic. If you want to expose certain endpoints, such as the /clients endpoint for OpenID Connect Dynamic Client Registry, you can do so but you need to properly secure these endpoints with an API Gateway or Authorization Proxy.```

And also:
``` None of the administrative endpoints have any built-in access control. You can do simple curl or Postman requests to talk to them.

The Token Introspection endpoint requires authentication. But since there is no access control, any valid authentication enables the endpoint to be used. If you need to access this endpoint in production, you should configure your API Gateway or Application Proxy to restrict which clients have access to the endpoint.```

Moreover, I couldn't find anything in my local setup nor in the configuration page that would suggest that hydra handles  authentication/authorization over the admin APIs.

If we need here an API key, I don't know anything about it. How will we host hydra? What kind of setup are we talking about here?

File: src/main/java/io/supertokens/webserver/api/oauth/OAuthAuthAPI.java (hunk 54 above)
Comment (tamassoltesz) [reply src/main/java/io/supertokens/webserver/api/oauth/OAuthAuthAPI.java:63]: Well, based on how the other InputParser methods are working, I would say no. The other inputParser methods are not changing the given input.
Regarding the aspect of sending a possible wrong value to hydra: I think hydra should handle it (and for example: sending a non-existent clientId will result in an error from hydra).
I don't think we should do any - possibly unexpected - changes to the input. In my opinion, we only should "translate" the property keys, to be able to communicate with hydra, and that's it. If the sdk gives a wrong input that's an error on the sdk part and the core handles it accordingly.

File: src/main/java/io/supertokens/webserver/InputParser.java (hunk 52 above)
Comment (tamassoltesz) [reply src/main/java/io/supertokens/webserver/InputParser.java:242]: insted of `Else` I use the `Or` in the fix, just to follow the naming which already exists in the InputParser

File: src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java [hunk 58]
```diff
@@ -76,8 +76,8 @@ protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws I
         } catch (OAuthClientRegisterInvalidInputException registerException) {
 
             JsonObject errorResponse = createJsonFromException(registerException);
-            errorResponse.addProperty("status", "INVALID_INPUT");
-            sendJsonResponse(200, errorResponse, resp);
+            errorResponse.addProperty("status", "INVALID_INPUT_ERROR");
+            sendJsonResponse(400, errorResponse, resp);
```
rishabhpoddar, line 80: we do this using the BadRequestException class that we have. The message in that can have the details of what's wrong with the input.

src/main/java/io/supertokens/oauth/OAuth.java
rishabhpoddar: It's ok. We can make it so that the admin endpoints are not exposed.

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
tamassoltesz: Okay. Done it.

rishabhpoddar (approved):

PR #1020, opened 2024-08-05
Feature: oauth update client api

tamassoltesz:
## Summary of change

Introduces the PATCH `/recipe/oauth/clients` endpoint for updating oauth2 clients. Also, adds PATCH method handling for the core.

## Related issues
[Make SuperTokens an OAuth and Open ID provider](https://github.com/supertokens/supertokens-core/issues/582)

## Documentation changes

    - In `build.gradle`

- [ ] Cover the sendJsonPATCHRequest method with tests


src/main/java/io/supertokens/httpRequest/HttpRequest.java
```diff
@@ -263,6 +262,31 @@ public static <T> T sendJsonPUTRequest(Main main, String requestID, String url,
         return sendJsonRequest(main, requestID, url, requestBody, connectionTimeoutMS, readTimeoutMS, version, "PUT");
     }
 
+    //TODO: tests!
```
rishabhpoddar, line 265: TODO here

src/main/java/io/supertokens/webserver/InputParser.java
```diff
@@ -239,7 +239,7 @@ public static Integer parseIntOrThrowError(JsonObject element, String fieldName,
 
     }
 
-    public static List<String> collectAllMissingRequiredFieldsOrThrowError(JsonObject input, List<String> requiredFields)
+    public static void collectAllMissingRequiredFieldsOrThrowError(JsonObject input, List<String> requiredFields)
```
rishabhpoddar, line 242: this function needs to be renamed given that we aren';t returning anything from it

src/main/java/io/supertokens/webserver/WebserverAPI.java
```diff
@@ -486,7 +490,12 @@ protected void service(HttpServletRequest req, HttpServletResponse resp) throws
                 Logging.info(main, tenantIdentifier,
                         "API called: " + req.getRequestURI() + ". Method: " + req.getMethod(), false);
             }
-            super.service(req, resp);
+            String method = req.getMethod();
+            if(method.equals("PATCH")){
```
rishabhpoddar, line 494: are we sure that it will always return all capital letters and not something like `patch` ever?

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
```diff
@@ -137,11 +139,46 @@ protected void doDelete(HttpServletRequest req, HttpServletResponse resp) throws
...
+
+            JsonObject response = OAuth.updateOauthClient(super.main, appIdentifier, storage, input);
+            JsonObject postResponseBody = new JsonObject();
+            postResponseBody.addProperty("status", "OK");
+            postResponseBody.add("client", response);
+            sendJsonResponse(200, postResponseBody, resp);
+
+        } catch (OAuthAPIInvalidInputException exception) {
+            JsonObject errorResponse = createJsonFromException(exception, INVALID_INPUT_ERROR);
+            sendJsonResponse(400, errorResponse, resp);
```
rishabhpoddar, line 159: we use the BadRequestException for this and the string response should contain the details. No need for status and stuff here

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
```diff
@@ -137,11 +139,46 @@ protected void doDelete(HttpServletRequest req, HttpServletResponse resp) throws
...
+
+        } catch (OAuthClientUpdateException updateException) {
+            //for errors with the update from hydra, which are not invalid input errors
+            JsonObject errorResponse = createJsonFromException(updateException, OAUTH2_CLIENT_UPDATE_ERROR);
+            sendJsonResponse(200, errorResponse, resp);
```
rishabhpoddar, line 164: this should result in a 500 error from the core as well - cause i suspect hydra will only return 500 in case something like the db is down or something?

rishabhpoddar (changes requested):

src/main/java/io/supertokens/httpRequest/HttpRequest.java
tamassoltesz: yeah. And also in the PR description. Should I remove it from the code?

src/main/java/io/supertokens/webserver/InputParser.java
tamassoltesz: done

src/main/java/io/supertokens/webserver/WebserverAPI.java
tamassoltesz: yes. I exactly followed the implementation in the tomcat HttpServletRequest with the other - supported - methods. Also, [RFC 2616](https://www.ietf.org/rfc/rfc2616.txt) states, that Method names are case-sensitive (section 5.1.1)

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
tamassoltesz: At this endpoint hydra's behaviour doesn't seem to be consistent. It can also return a 500 for example if the input json is empty. (So, not just for only real internal server errors, but for like programmer errors and stuff too.. I couldn't get a guideline out of hydra about this error handling..)
I can change the return code to 500 in the core also, if you like, but I don't think that status code carries the message of what went wrong. (But of course, we can say that "hydra went wrong", which is an internal error - only it went wrong because a wrong input)

src/main/java/io/supertokens/httpRequest/HttpRequest.java
rishabhpoddar: I didn't understand your comment.

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
rishabhpoddar: We should send a 500, with a generic error - we should not expose to the user that we are using hydra

tamassoltesz: Okay. The error hydra sends back in these cases is: `{error: "error", error_description: "The error is unrecognizable"}`. I would suggest to go with the more generic "Internal server error". Wdyt?

src/main/java/io/supertokens/httpRequest/HttpRequest.java
rishabhpoddar: Oh yea, please do. It can be added in the checklist here: https://github.com/supertokens/supertokens-core/pull/1010

src/main/java/io/supertokens/webserver/api/oauth/OAuthClientsAPI.java
rishabhpoddar: Sure.

src/main/java/io/supertokens/httpRequest/HttpRequest.java
tamassoltesz: okay, done

Issue #1048, opened 2024-09-26
The docker image requires root permission and is unable to run on a read only file system


rishabhpoddar:
## 🐛 Bug Report

At the moment, the user in the docker image requires root permission to run, and the docker image itself is unable to run on a read only root file system. This poses security risks that need to be dealt with.

mavridiSS:
+1 - This issue came up on a pen test report and we're waiting for a fix for it to go live.

@tamassoltesz do you have any rough estimations?

MoggyPersonal:
+1 - Will definitely be flagged within our own security reports that clients ask for

Would be great to know progress on this please

rishabhpoddar:
We are working on this now and should have a fix for it sometime next week.

sattvikc:
Fixed in core version 9.2.3

tamassoltesz:
Root permission/root user is no longer needed in the container.
For --read-only (read-only root filesystem), you should mount a tmpfs to `/lib/supertokens/temp`  with `:exec` mode. (see [https://github.com/supertokens/supertokens-docker-postgresql/tree/9.2 README](https://github.com/supertokens/supertokens-docker-postgresql/blob/9.2/README.md#read-only-root-fs))

PR #1058, opened 2024-10-08
fix: docker readonly fs

tamassoltesz:
## Summary of change

Adds support in CLI and in Core to specify the temp folder location. It's required for docker when run with readonly root fs.

## Related issues

[1048](https://github.com/supertokens/supertokens-core/issues/1048)

## Test Plan

## Documentation changes

    - In `build.gradle`

### DIFF
--- a/src/main/java/io/supertokens/Main.java
+++ b/src/main/java/io/supertokens/Main.java
@@ -266,6 +266,7 @@ public class Main {
         Webserver.getInstance(this).start();
 
         // this is a sign to the controlling script that this process has started.
+
         createDotStartedFileForThisProcess();
 
         // NOTE: If the message below is changed, make sure to also change the corresponding check in the CLI program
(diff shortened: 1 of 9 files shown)


src/main/java/io/supertokens/Main.java
```diff
@@ -345,10 +346,11 @@ public void proceedToEnableFeatureFlag() {
 
     private void createDotStartedFileForThisProcess() throws IOException {
         CoreConfig config = Config.getBaseConfig(this);
+        String fileLocation = CLIOptions.get(this).getTempDirLocation() == null ? CLIOptions.get(this).getInstallationPath() : CLIOptions.get(this).getTempDirLocation();
```
sattvikc, line 349: Not related to this file. CoreConfig is also using getInstallationPath() to figure out path for config.yaml and log file paths. Please see all places where we might finding a directory to write to.

sattvikc (changes requested):

src/main/java/io/supertokens/Main.java
tamassoltesz: This is a special case for docker readonly fs. We are not intending to overwrite the installation path, or any of it. In the places where those are used, if those values (so, config location, log location) are not set, then it's a misconfigured core. The docker entrypoint sets those values as required. 
Moreover, If we automatically set the config location to the temp-dir-location, that could cause issues for example when the user intends to provide his/her own config.yml and also set a temp-dir-location (in docker)

sattvikc (approved):

Issue #319, opened 2021-10-04
System.err.println(err);  needs possible redirect 


spsystemsAFK:
https://github.com/supertokens/supertokens-core/blob/aaa0628bdf266308c349e905376a8efb90c30a6f/src/main/java/io/supertokens/output/Logging.java#L147

rishabhpoddar:
Related to https://github.com/supertokens/supertokens-core/issues/5

Issue #5, opened 2020-09-11
Docker image's error log not in stderr


rishabhpoddar:

AnmolVirdi:
Could you please elaborate the issue. I'm willing to work on the Dockerfile

rishabhpoddar:
@AnmolVirdi , the core has two types of outputs:
1) Debug / info logs
2) Error logs

Without docker, the output of (1) is sent to a file called `info.log`, and output of (2) is sent to `error.log`. In the code, the error logs are printed using `System.err.println`, [here](https://github.com/supertokens/supertokens-core/blob/master/src/main/java/io/supertokens/output/Logging.java#L147) and the info logs are printed using `System.out.println`, [here](https://github.com/supertokens/supertokens-core/blob/master/src/main/java/io/supertokens/output/Logging.java#L142).

The problem comes in the docker container, where both these types of logs are being printer to the `STDOUT` instead of one being printed to `STDOUT` and one to `STDERROR`.

yowayb:
Has anyone been able to reproduce this?

I forced an error by putting an exclamation mark before `core_config_version` in `config.yaml` and colored  `stderr` red:

![image.png](https://images.zenhubusercontent.com/61e1953204116bee3a23e463/19b6dc41-db95-4747-891c-a08e1a94210f)

It appears to work.

rishabhpoddar:
Okay thanks. We will recheck this

rishabhpoddar:
It seems that `https://github.com/supertokens/supertokens-core/blob/master/src/main/java/io/supertokens/webserver/WebserverAPI.java#L161` and `https://github.com/supertokens/supertokens-core/blob/master/src/main/java/io/supertokens/webserver/WebserverAPI.java#L179` don't print out on stderr which is perhaps the root of the issue here.

relentless-pursuit:
@rishabhpoddar, is the issue still relevant? Can I take it up?

ADITYA-CODE-SOURCE:
@rishabhpoddar i’d be happy to take this on. Before opening a PR, I want to quickly verify whether the issue still reproduces on the current Docker setup, since the logging/CLI behavior may have changed since this was opened. If it does, I’ll trace the exact codepath and put up a targeted fix.

PR #1245, opened 2026-03-24
fix(cli): keep child process stderr separate from stdout in StartHandler

sakthiveltofficial:
Summary of change
This PR fixes log stream handling when starting SuperTokens via the CLI. Previously, StartHandler merged stderr into stdout using redirectErrorStream(true), which caused error logs to be surfaced as info logs.
Now, stdout and stderr are handled separately: stdout is still used for startup detection, while stderr is read independently and logged via Logging.error, preserving proper error visibility.

Related issues
https://github.com/supertokens/supertokens-core/issues/5

https://github.com/supertokens/supertokens-core/issues/319

Test Plan
Ran CLI unit/integration tests:

mvn -pl cli -DskipTests=false test

Verified that tests pass successfully.

Manual verification:

Start SuperTokens using CLI start command.

Trigger an error condition in core startup/runtime.

Confirm normal startup/info messages appear on stdout.

Confirm error output from the child process appears on stderr and is logged as errors.

Documentation changes
No documentation updates are required for this internal logging/stream-handling fix.

Checklist for important updates
 Changelog has been updated

 If there are any db schema changes, mention those changes clearly

 coreDriverInterfaceSupported.json file has been updated (if needed)

 pluginInterfaceSupported.json file has been updated (if needed)

 Changes to the version if needed

In build.gradle

 If added a new paid feature, edit the getPaidFeatureStats function in FeatureFlag.java file

 Had installed and ran the pre-commit hook

 If there are new dependencies that have been added in build.gradle, please make sure to add them
in implementationDependencies.json.

 Update function getValidFields in io/supertokens/config/CoreConfig.java if new aliases were added for any core
config (similar to the access_token_signing_key_update_interval config alias).

 Issue this PR against the latest non released version branch.

To know which one it is, run find the latest released tag (git tag) in the format vX.Y.Z, and then find the
latest branch (git branch --all) whose X.Y is greater than the latest released tag.

If no such branch exists, then create one from the latest released branch.

 If added a foreign key constraint on app_id_to_user_id table, make sure to delete from this table when deleting
the user as well if deleteUserIdMappingToo is false.

 If added a new recipe, then make sure to update the bulk import API to include the new recipe.

### DIFF
--- a/cli/src/main/java/io/supertokens/cli/commandHandler/start/StartHandler.java
+++ b/cli/src/main/java/io/supertokens/cli/commandHandler/start/StartHandler.java
@@ -25,6 +25,7 @@ import io.supertokens.cli.logging.Logging;
 
 import java.io.BufferedReader;
 import java.io.File;
+import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


cli/src/main/java/io/supertokens/cli/commandHandler/start/StartHandler.java
```diff
@@ -191,6 +197,22 @@ public void doCommand(String installationDir, boolean viaInstaller, String[] arg
         }
     }
 
+    private static void logProcessStream(InputStream inputStream, boolean isErrorStream) {
```
tamassoltesz, line 200: Is this ever called with isErrorStream = false? (Seems like not.) Please don't solve _possible_ future problems, only solve the problem at hand.

tamassoltesz (changes requested):

PR #1275, opened 2026-05-21
feat: backfill api

tamassoltesz:
## Summary of change

## Related issues

- Link to issue1 here
- Link to issue1 here

## Test Plan

## Documentation changes

    - In `build.gradle`

- [ ] Item1
- [ ] Item2

### DIFF
--- a/src/main/java/io/supertokens/inmemorydb/Start.java
+++ b/src/main/java/io/supertokens/inmemorydb/Start.java
@@ -4297,6 +4297,11 @@ public class Start
     }
 
     @Override
+    public void setMigrationMode(MigrationMode mode) {
+        SQLiteConfig.setMigrationModeForTesting(mode);
+    }
+
+    @Override
     public int getBackfillPendingUsersCount(AppIdentifier appIdentifier) {
         return 0; // InMemoryDB always has all data in sync
     }
--- a/src/main/java/io/supertokens/inmemorydb/Start.java
+++ b/src/main/java/io/supertokens/inmemorydb/Start.java
@@ -4283,11 +4283,6 @@ public class Start
     }
 
     @Override
-    public void setMigrationMode(MigrationMode mode) {
-        SQLiteConfig.setMigrationMode(mode);
-    }
-
-    @Override
     public int getBackfillPendingUsersCount(AppIdentifier appIdentifier) {
         return 0; // InMemoryDB always has all data in sync
     }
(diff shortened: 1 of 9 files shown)


src/main/java/io/supertokens/inmemorydb/Start.java
```diff
@@ -4270,6 +4270,11 @@ public MigrationMode getMigrationMode() {
         return Config.getConfig(this).getMigrationMode();
     }
 
+    @Override
+    public void setMigrationMode(MigrationMode mode) {
```
porcellus: Why do we need a separate function for this?

src/main/java/io/supertokens/webserver/api/migration/MigrationBackfillProgressAPI.java
```diff
@@ -0,0 +1,113 @@
...
+        } catch (TenantOrAppNotFoundException | StorageQueryException e) {
+            throw new ServletException(e);
+        }
+    }
+
+    private JsonObject buildProgressEntry(MigrationBackfillStorage mbs, AppIdentifier app)
+            throws StorageQueryException {
+        JsonObject entry = new JsonObject();
+        entry.addProperty("mode", mbs.getMigrationMode().name());
+        int pending = mbs.getBackfillPendingUsersCount(app);
```
porcellus: I think we can shortcut this for all modes except read old

src/main/java/io/supertokens/webserver/api/migration/MigrationBackfillProgressAPI.java
```diff
@@ -0,0 +1,113 @@
...
+        entry.addProperty("pendingUsers", pending);
+        // Defer the expensive full-scan completeness check until the backfill is done.
```
porcellus: We have to shortcut this for already migrated cuds.
We may also want to put this behind a flag so it's only checked right before we intend to flip the migration mode.

porcellus (changes requested):

src/main/java/io/supertokens/inmemorydb/Start.java
tamassoltesz: `setMigrationMode` is called by `MigrationModeAPI.doPut` through the `MigrationBackfillStorage` interface — it's the runtime API that lets operators flip the migration mode. The in-memory DB implements it by delegating to `SQLiteConfig.setMigrationModeForTesting` since that's the existing knob for changing runtime config in tests. (Also, that's right that the method is called ...`Testing`, but without a @TestOnly annotation, so maybe it would be better to get rid of that `..Testing` method all together, and keep only this one)

porcellus (approved):

PR #1313, opened 2026-07-27
fix: prevent duplicate access token signing keys on concurrent rotation

mchoopani:
## The problem

When multiple core instances run behind a load balancer and the dynamic access token signing key comes up for rotation, each instance races to create the new key and more than one can succeed. Each instance then caches and signs tokens with its own key while serving a JWKS that only contains that key. Consumers validating against the JWKS (Envoy's `jwt_authn`, for example) intermittently reject valid tokens with `Jwks_doesn't_have_key_to_match_kid_or_alg_from_Jwt` until the next rotation happens to reconverge the caches.

The root cause is in `AccessTokenSigningKey.getOrCreateAndGetSigningKeys()`, which does a check-then-act (`SELECT ... FOR UPDATE`, then `INSERT` if no fresh key exists) at the default `READ COMMITTED` isolation level. `FOR UPDATE` only locks rows that already exist, and under `READ COMMITTED` a blocked `SELECT` re-reads only the rows it was blocked on — not rows inserted by the transaction it was waiting for. So a second instance that blocks on the first instance's lock still doesn't see the first instance's newly committed key when it unblocks. Both instances conclude that no fresh key exists and both insert one. Since the primary key includes `created_at_time` and the two inserts land on different milliseconds, there's no unique violation and both commits succeed silently. The existing `startTransaction` retry wrapper never kicks in because nothing fails — there's no deadlock, no constraint violation, and no serialization error at this isolation level.

## Reproducing with two psql sessions

You can demonstrate the race directly against the `session_access_token_signing_keys` table (assuming at least one existing key row so there's something to lock):

```sql
-- Session A                                    -- Session B
BEGIN;
SELECT * FROM session_access_token_signing_keys
  WHERE app_id = 'public' FOR UPDATE;
-- sees only the old, stale key
                                                BEGIN;
                                                SELECT * FROM session_access_token_signing_keys
                                                  WHERE app_id = 'public' FOR UPDATE;
                                                -- blocks on A's row locks...

INSERT INTO session_access_token_signing_keys
  VALUES ('public', 1700000000001, 'keyA...');
COMMIT;
                                                -- ...unblocks, but the result set
                                                -- still does NOT include A's new key,
                                                -- so B also decides to create one
                                                INSERT INTO session_access_token_signing_keys
                                                  VALUES ('public', 1700000000002, 'keyB...');
                                                COMMIT;  -- succeeds silently
```

The table now has two keys for a single rotation. If you repeat the same steps but start both sessions with `BEGIN ISOLATION LEVEL SERIALIZABLE`, session B's commit instead fails with `serialization_failure` (SQLSTATE `40001`) — which is exactly what this fix relies on.

## The fix

Run that one transaction at `SERIALIZABLE` isolation:

```java
validKeys = sqlStorage.startTransaction(con -> { ... }, TransactionIsolationLevel.SERIALIZABLE);
```

Under serializable snapshot isolation the two concurrent creators form a read/write conflict, so the second one to commit aborts with SQLSTATE `40001`. The PostgreSQL plugin's existing `startTransaction` retry loop already catches class-40 errors and re-runs the transaction; on retry the aborted instance takes a fresh snapshot, sees the other instance's committed key, and skips its own insert. Exactly one key is created per rotation. This mirrors the approach taken in #1257 for the analogous OAuth refresh-token race.

The change is core-only and uses the existing `startTransaction(logic, isolationLevel)` overload, so no plugin interface changes are needed. The PostgreSQL plugin already honours the requested isolation level and retries serialization failures. The in-memory (SQLite) plugin ignores the isolation level, but it is already protected by its per-app connection locking, so passing `SERIALIZABLE` is a harmless no-op there. The NoSQL path is already safe through its optimistic-concurrency retry loop.

## Notes

The static JWT-recipe key path (`JWTSigningKey.getOrCreateAndGetKeyForAlgorithm`) has the same check-then-act pattern, but since it isn't involved in access-token/JWKS validation it is intentionally left out to keep this PR focused; happy to follow up separately. Deployments already affected by this race may also want to clean up duplicate rows (keep the newest `created_at_time` per `app_id`), since instance caches will keep diverging on the stale duplicates until the next rotation.

## Changelog

```markdown
## [Unreleased]

- Fixes a race that could create duplicate access token signing keys when multiple
  core instances rotated the key concurrently (SQL/PostgreSQL storage). Key creation
  now runs at SERIALIZABLE isolation, so a concurrent creator aborts with a
  serialization failure and is retried by the existing transaction-retry logic
  instead of inserting a second key. Duplicate keys caused JWKS `kid` mismatches
  during rotation (e.g. Envoy `Jwks_doesn't_have_key_to_match_kid_or_alg_from_Jwt`).
```

### DIFF
--- a/src/main/java/io/supertokens/signingkeys/AccessTokenSigningKey.java
+++ b/src/main/java/io/supertokens/signingkeys/AccessTokenSigningKey.java
@@ -39,6 +39,7 @@ import io.supertokens.pluginInterface.multitenancy.exceptions.TenantOrAppNotFoun
 import io.supertokens.pluginInterface.session.SessionStorage;
 import io.supertokens.pluginInterface.session.noSqlStorage.SessionNoSQLStorage_1;
 import io.supertokens.pluginInterface.session.sqlStorage.SessionSQLStorage;
+import io.supertokens.pluginInterface.sqlStorage.SQLStorage.TransactionIsolationLevel;
 import io.supertokens.storageLayer.StorageLayer;
 import io.supertokens.utils.Utils;
 import org.jetbrains.annotations.TestOnly;
@@ -226,7 +227,10 @@ public class AccessTokenSigningKey extends ResourceDistributor.SingletonResource
         if (storage.getType() == STORAGE_TYPE.SQL) {
             SessionSQLStorage sqlStorage = (SessionSQLStorage) storage;
             try {
-                // start transaction
+                // SERIALIZABLE (not the default READ_COMMITTED): on a concurrent rotation, READ COMMITTED lets
+                // each instance's SELECT snapshot predate the other's INSERT, so both create a key. SERIALIZABLE
+                // aborts the loser with serialization_failure (SQLSTATE 40); the plugin's startTransaction retry
+                // re-runs it, by which time the other's key is visible, so no duplicate is created.
                 validKeys = sqlStorage.startTransaction(con -> {
                     List<SigningKeys.KeyInfo> validKeysFromSQL = new ArrayList<>();
 
@@ -269,7 +273,7 @@ public class AccessTokenSigningKey extends ResourceDistributor.SingletonResource
 
                     sqlStorage.commitTransaction(con);
                     return validKeysFromSQL;
-                });
+                }, TransactionIsolationLevel.SERIALIZABLE);
             } catch (StorageTransactionLogicException e) {
                 if (e.actualException instanceof TenantOrAppNotFoundException) {
                     throw (TenantOrAppNotFoundException) e.actualException;


src/main/java/io/supertokens/signingkeys/AccessTokenSigningKey.java
```diff
@@ -269,7 +273,7 @@ public List<SigningKeys.KeyInfo> getOrCreateAndGetSigningKeys()
 
                     sqlStorage.commitTransaction(con);
                     return validKeysFromSQL;
-                });
+                }, TransactionIsolationLevel.SERIALIZABLE);
```
tamassoltesz, line 276: overall, I'm not happy with this. Changing isolation level per transaction could and will lead to lock accumulation - which was a major cause for a pretty big outage for us. The whole 12.0 release's main goal is to move away from 
1. SERIALIZABLE isolation level
2. changing the isolation level per transaction.

I'm not yet sure how, but I'm sure there is an other solution for the problem.

tamassoltesz (changes requested):

src/main/java/io/supertokens/signingkeys/AccessTokenSigningKey.java
mchoopani: I understand your concern — the SERIALIZABLE approach doesn't align well with the direction of 12.0, so let's avoid changing transaction isolation here.

One alternative that doesn't require that is an **advisory lock** scoped by `app_id`. This would serialize key creation for a given app and doesn't depend on any existing rows being present, so it also handles the first-key-creation case cleanly.

Another option would be to enforce the invariant at the schema level with additional constraints. That's likely the strongest guarantee from a correctness perspective; I was mainly considering the advisory-lock approach as a smaller, more localized change. Happy to go with whichever direction makes the most sense.

tamassoltesz: How would you enforce the invariant at the schema level?

mchoopani: Two options:

1. Add an is_current flag with a unique constraint per app_id, ensuring only one active signing key. Concurrent rotations would conflict on the constraint and require retry.
2. Model key lineage (e.g., store predecessor key hash) and enforce a single rotation per key via a unique constraint (unique on (app_id, prev_key_hash)). Concurrent rotations from the same key would conflict on insert.

mchoopani: Thanks for your feedback so far. I'd love to resolve the remaining points and get this PR wrapped up efficiently so we can fix this concurrent key rotation bug for the community.
If you prefer, please let me know if there's a better channel to discuss this more quickly.

tamassoltesz: Hey, Thanks for your patience! I would suggest to go with the advisory lock. Also, ideally, you would create a test that reproduces the issue, then fix the issue (so the test goes "green"), locking the expected behavior with a test.

mchoopani: Hey! Thanks again for the guidance — I've gone with the advisory lock approach as suggested and added a test that reproduces the race first.

Split it across two PRs:

[postgresql-plugin](https://github.com/supertokens/supertokens-postgresql-plugin/pull/386)
[core](https://github.com/supertokens/supertokens-core/pull/1389)

PR #1383, opened 2026-08-14
ci: automate migration-script awareness at core releases

tamassoltesz:
## Summary of change

Automates *DB-migration awareness* into the core release pipeline, so that a core release which ships schema changes safely defers the SaaS (ECS) rollout until the change is known on the backend-apis side.

`has-db-migration` is **derived, not hand-typed**: the postgresql plugin's `migration-scripts/manifest.json`, read at the exact plugin SHA being released, is the single source of truth. When a release ships a migration:

- the core's own `CHANGELOG.md` is required to document it (checked early via the plugin's `schema-migration-check.py`; the release stops before anything is registered if it doesn't);
- the derived flag is sent on the `PATCH /core` API calls (`hasDBMigration`) so the SaaS backend can defer its rollout with an alert + pending marker;
- a PR is opened against `supertokens-backend-apis` adding the migration to the `CORE_MIGRATIONS` registry (the change the rollout guard waits for). Idempotent; `continue-on-error` so PR creation never fails the release.

Absent/old manifests derive `false`, so older plugin versions and backports never trigger any of this.

### Cross-repo companion (must release together)

- **supertokens-root**, branch `feat/automate_migration_awareness_at_core_release` (commit 6745163): `utils/addReleaseTag` now reads the `HAS_DB_MIGRATION` env this workflow exports (`hasDBMigration=${HAS_DB_MIGRATION:-false}`) and forwards it on the core PATCH. `loadModules` copies that script into the core module dir at release time — it is the consumer of the env wired into the `Tag core` step.
- **supertokens-backend-apis**: receiver of the auto-generated migration PRs (`CORE_MIGRATIONS` registry + `saas_rollout_pending` guard).

## Test Plan

CI-only change (GitHub Actions workflows). Validated the changed/added workflow YAML parses and that the reusable-workflow inputs/outputs plumbing matches between callers and `_release-derive-db-migration.yml`. The release paths themselves run only on `workflow_dispatch` at release time.

## Review-round changes (addressing @porcellus and the automated review)

- **Canary excluded** from `open-backend-apis-migration-pr` (`&& !inputs.is-canary`) — a canary shares the non-canary `core-version`, so it could otherwise open a prod backend-apis PR and pre-empt the real release via the branch-exists guard.
- **Deduplicated** the release-critical derive logic into a reusable workflow `_release-derive-db-migration.yml`, shared by `do-release.yml` and `do-release-java15.yml`.
- **Robust JSON**: java15 `mark-as-passed` now normalizes the flag with `== 'true'` (no invalid-JSON path on a blank output).
- **Injection hardening**: release-version inputs pass through `env:` in the backend-apis PR step instead of being expanded into the shell.
- The java15 (backport) path intentionally has no backend-apis PR job — a backport that ships a migration still defers the SaaS rollout via the flag; the registry entry is added by hand (documented inline).

## Documentation changes

CONTRIBUTING.md / README updated to describe the migration-awareness pipeline (in this PR).


.github/workflows/do-release.yml
```diff
@@ -85,8 +85,75 @@ jobs:
...
+          path: core
+      - name: Derive has-db-migration
+        id: derive
+        run: |
+          manifest="plugin/migration-scripts/manifest.json"
+          pluginVersion="${{ inputs.postgresql-plugin-version }}"
+          if [ -f "$manifest" ]; then
+            entry=$(jq -c --arg v "$pluginVersion" '.migrations[] | select(.pluginVersion == $v)' "$manifest")
+          else
+            echo "::notice::$manifest not present at this postgresql-plugin SHA - assuming no DB migration."
```
porcellus: Maybe we should only allow this if we have an opt-in flag. This would only be the case for older plugin versions, right (so backport only)?

porcellus (changes requested):

.github/workflows/do-release.yml
tamassoltesz: No opt-in flag needed — it's already safe-by-default, and it's not backport-only.

`has-db-migration` is derived solely from the postgresql plugin's `migration-scripts/manifest.json`, read at the exact plugin SHA being released (now in the shared `_release-derive-db-migration.yml`). So:

- If that plugin version has a manifest entry for the release → the migration pipeline runs. This applies to **any** release that ships a schema change, not just backports.
- If there's no entry, or the manifest predates the convention entirely (the `else` branch) → it derives `false` and the whole backend-apis/SaaS-defer pipeline stays off.

That means a release only ever "opts in" by the plugin actually shipping a migration script — there's no way for an older plugin version or a backport to trigger it accidentally, which is the guarantee an explicit opt-in flag would otherwise buy us. Adding a manual flag on top would just duplicate the manifest signal and introduce a way to forget it on a real migration release (the failure mode this automation exists to remove).

The canary case you'd worry about is handled separately: `open-backend-apis-migration-pr` is now gated with `&& !inputs.is-canary`, so a canary can't open a prod backend-apis PR or pre-empt the real release's branch-exists guard.

Could you take another look when you have a moment?

.github/workflows/do-release.yml
```diff
@@ -108,16 +175,133 @@ jobs:
...
+  # guard-satisfying-but-wrong entry is worse than a blocking one. Everything
+  # is keyed by the full core X.Y.Z (patch releases can ship index-only
+  # migrations too). Skipped when the registry already has the version or the
+  # PR branch already exists, so re-releases stay idempotent. continue-on-error: PR creation must never
+  # fail the release — the rollout guard + pending marker + Slack reminder
+  # still cover the case where this job breaks.
+  open-backend-apis-migration-pr:
+    name: Open backend-apis migration PR
+    needs: [ dependency-branches, derive-db-migration-flag, mark-as-passed ]
+    if: ${{ !cancelled() && !failure() && needs.derive-db-migration-flag.outputs.has-db-migration == 'true' }}
```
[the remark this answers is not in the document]
tamassoltesz: actually, we want this to happen on canary relases too. Please remove the canary-gate

PR #1385, opened 2026-08-19
fix: bulk import no longer borrows from the live connection pool

tamassoltesz:
The cron checks for work on the live pool, then opens one dedicated pool per user pool of the app, sized to bulk_migration_parallelism. Each worker claims its chunk (FOR UPDATE SKIP LOCKED), imports the users and deletes or error-marks the rows on a single connection from that pool, in one transaction; a failed partition rolls back to a savepoint so the claim is kept until the rows reach a terminal state, which keeps multiple core instances on the same database from importing a user twice. Per-partition deletes, bounded immediate retries, and BulkImport.importUser use the same mechanism.

## Summary of change

## Related issues

- Link to issue1 here
- Link to issue1 here

## Test Plan

## Documentation changes

    - In `build.gradle`

- [ ] Item1
- [ ] Item2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,22 +7,20 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+## [12.2.0]
+
+- **Upgrade note: the core now verifies the database schema at startup and, by default (`schema_check_strict_mode: true`), refuses to start when the base database is missing a manual migration — run the migration SQL from the CHANGELOGs (or set `schema_check_strict_mode: false`) before upgrading**
+- Verifies the database schema at startup: any database (base or tenant) missing a table or column this version needs is reported with the missing columns plus the SQL to add them
+- New config `schema_check_strict_mode` (boolean, default `true`, config.yaml / env only): in strict mode a mismatched tenant database refuses all queries until the migration is applied — re-checked every minute, so it resumes within a minute of the migration, no restart needed; the core still boots and serves all other tenants
+- With `schema_check_strict_mode: false`, mismatches are only logged: everything keeps working and just the queries touching the missing schema fail, with a "Schema mismatch ... check the core error logs" hint instead of a raw SQL error
+- Corrects the 12.1.0 migration note: the `session_info` columns are a manual step, not applied automatically
+
 ## [12.1.2]
 
 - Bulk import no longer borrows from the live connection pool: each worker claims, imports and finalises its chunk on one connection from a dedicated pool sized to `bulk_migration_parallelism`
 - Bulk import keeps claimed `bulk_import_users` rows locked until they are deleted or error-marked; a failed chunk rolls back to a savepoint instead of releasing the claim
 - Bulk import deletes only the rows of the partition it imported and bounds immediate retries after a database rollback
 - `BulkImport.importUser` (the single-user import API) now runs on the same bounded dedicated pool, opened for the duration of the call (one connection per user pool, previously a full-size proxy pool per user pool)
-- Verifies the database schema at startup: any database (base or tenant) missing a table or column this version
-  needs is reported with the missing columns plus the SQL to add them
-- **New config `schema_check_strict_mode` (boolean, default `true`, config.yaml / env only). In strict mode the
-  core REFUSES TO START when the base database schema is out of date (a skipped manual migration), and tenant
-  databases with a mismatched schema are not served until the migration is applied. If your database schema is
-  behind (see the Migration sections below), upgrading to this version will stop the core from booting until you
-  either run the migration SQL or set `schema_check_strict_mode: false`.** With strict mode off, mismatches are
-  only logged: the core boots, everything else keeps working, and just the queries touching the missing schema
-  fail - with a "Schema mismatch ... check the core error logs" message instead of a raw SQL error
-- Corrects the 12.1.0 migration note: the `session_info` columns are a manual step, not applied automatically
 
 ## [12.1.1]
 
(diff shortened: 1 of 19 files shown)


src/main/java/io/supertokens/bulkimport/BulkImport.java
```diff
@@ -176,16 +170,17 @@ public static synchronized AuthRecipeUserInfo importUser(Main main, AppIdentifie
...
                 appIdentifier.getAppId(), user.loginMethods.get(0).tenantIds.get(0));
 
-        SQLStorage bulkImportProxyStorage = (SQLStorage) getBulkImportProxyStorage(main, firstTenantIdentifier);
-
         LoginMethod primaryLM = BulkImportUserUtils.getPrimaryLoginMethod(user);
 
-        try {
+        // A single-connection import: one proxy storage per user pool of the app, each on its own dedicated
+        // one-connection pool, all released when done.
+        try (BulkImportProxyStoragePools pools = BulkImportProxyStoragePools.openForApp(main, appIdentifier, 1);
```
porcellus, line 177: Since this is called directly via an API, is there a risk it will flood the DB with connections?

src/main/java/io/supertokens/cronjobs/bulkimport/ProcessBulkUsersImportWorker.java
```diff
@@ -272,72 +268,13 @@ private static void handleBulkImportException(List<BulkImportUser> usersBatch,
...
         for (BulkImportUser user : users) {
-            TenantIdentifier firstTenantIdentifier = new TenantIdentifier(appIdentifier.getConnectionUriDomain(),
-                    appIdentifier.getAppId(), user.loginMethods.getFirst().tenantIds.getFirst());
-            
-            SQLStorage bulkImportProxyStorage = (SQLStorage) getBulkImportProxyStorage(firstTenantIdentifier);
-            if (!result.containsKey(bulkImportProxyStorage)) {
-                result.put(bulkImportProxyStorage, new ArrayList<>());
-            }
-            result.get(bulkImportProxyStorage).add(user);
+            TenantIdentifier firstTenantIdentifier = new TenantIdentifier(app.getConnectionUriDomain(),
```
porcellus, line 275: Are the tenant IDs ordered in any way? Do we care?

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+## [12.1.2]
+
+- Bulk import no longer borrows from the live connection pool: each worker claims, imports and finalises its chunk on one connection from a dedicated pool sized to `bulk_migration_parallelism`
+- Bulk import keeps claimed `bulk_import_users` rows locked until they are deleted or error-marked; a failed chunk rolls back to a savepoint instead of releasing the claim
+- Bulk import deletes only the rows of the partition it imported and bounds immediate retries after a database rollback
+- `BulkImport.importUser` (the single-user import API) uses the same dedicated-pool mechanism instead of a throwaway pool per call
```
porcellus: Isn't it still creating a throwaway pool?

src/main/java/io/supertokens/bulkimport/BulkImport.java
tamassoltesz: No — the exposure here is bounded, and smaller than before this PR. `importUser` is `static synchronized` (as it was previously), so a core instance runs at most one single-user import at a time; concurrent API calls queue on the lock instead of each opening pools. Within a call, the pools are opened with max 1 connection per distinct user pool of the app, `minimumIdle = 0`, and connections are borrowed lazily on first use — then everything is closed by the try-with-resources when the call returns. So the worst case at any moment is one connection per distinct database the app's tenants live on, per core instance; for a typical single-database app that's exactly 1.

Before this PR each call built a full proxy storage per user pool via `initStorage()` — a Hikari pool sized to the configured `connection_pool_size` (default 10) that also re-ran the startup table checks — so this is strictly a reduction. There is per-call pool open/close churn, but this endpoint is a migration aid rather than a hot path; if that ever changes we can cache the pool the same way the cron holds it for a run.

src/main/java/io/supertokens/cronjobs/bulkimport/ProcessBulkUsersImportWorker.java
tamassoltesz: They're in whatever order the operator supplied in the uploaded JSON — no guaranteed ordering, and we don't rely on one. What makes `getFirst().getFirst()` safe is the invariant validated when the entry is added (`BulkImportUserUtils`, "Now validate that all the tenants share the same storage"): every tenant across all login methods of one user must resolve to the same user pool, otherwise the add fails with "All tenants for a user must share the same database". Since `storages.forTenant()` maps tenant → userPoolId → storage, any tenant of any login method picks the same storage, so the choice is arbitrary but deterministic. This is the same convention `importUser` already used (see its comment "Since all the tenants of a user must share the storage..."). If storages were to diverge after validation (a tenant re-pointed at a different database before processing), the affected user's import fails and the row gets error-marked — it can't silently write into the wrong database.

CHANGELOG.md
tamassoltesz: Fair — it does still open a pool for the duration of the call; the line oversells it. What changed is what that pool is: previously each call created a full proxy storage per user pool via `initStorage()` (a Hikari pool sized to `connection_pool_size`, default 10, which also replayed the startup table checks); now it opens the same bounded `BulkImportConnectionPool` the cron uses — max 1 connection per user pool, no idle pre-fill, no DDL, tagged `ApplicationName=supertokens-bulk-import`, closed when the call returns. Reworded the entry.

Issue #1386, opened 2026-08-19
12.1.0 `session_info` columns are never added when upgrading an existing PostgreSQL database


JoseGoncalves:
## Summary

The 12.1.0 CHANGELOG says the new `session_info` columns are "additive only, applied
automatically at startup". They are not — there is no `ALTER TABLE` path in the postgresql
plugin. Upgrading any pre-12.1.0 deployment leaves `session_info` without
`prev_refresh_token_hash_2` / `refresh_token_rotated_at`, and every session read then fails.

Upgraded 12.0.10 → 12.1.1 (plugin interface 9.0), PostgreSQL, database created by an earlier
12.0.x core. `GET /recipe/session` returns 500:

```
org.postgresql.util.PSQLException: ERROR: column sess.prev_refresh_token_hash_2 does not exist
  at io.supertokens.storage.postgresql.queries.SessionQueries.getSession_impl(SessionQueries.java:509)
  at io.supertokens.storage.postgresql.queries.SessionQueries.getSession_legacy(SessionQueries.java:486)
  at io.supertokens.storage.postgresql.Start.getSession(Start.java:692)
  at io.supertokens.webserver.api.session.SessionAPI.doGet(SessionAPI.java:178)
```

The core boots cleanly and logs nothing about the skipped migration, so the first symptom is
a runtime 500 per request.

## Root cause

`SessionQueries.getQueryToCreateSessionInfoTable()` declares the columns, but only inside
`CREATE TABLE IF NOT EXISTS` — a no-op on an existing table. Neither `GeneralQueries.java`
nor `Start.java` contains any `ALTER TABLE` / `ADD COLUMN`, so nothing can add these columns
to a database created by an earlier version. Fresh installs are unaffected.

Not a permissions or environment problem: the configured role owns the table
(`session_info` / `supertokens_user`) and the core is in the expected database and schema, so
an `ALTER TABLE` would have succeeded had one been issued.

## Documentation discrepancy

- core 12.1.0: "Additive only, applied automatically at startup; no backfill required."
- postgresql-plugin 9.7.0: lists the same two `ALTER TABLE` statements under **Migration**,
  with no claim of automatic application.

The plugin's is correct. Operators following the core changelog have no reason to run the SQL
and get no warning that it was skipped.

## Workaround

```sql
ALTER TABLE session_info ADD COLUMN IF NOT EXISTS prev_refresh_token_hash_2 VARCHAR(128);
[...]

PR #1387, opened 2026-08-19
fix: verify database schema at startup and fail loudly on missing tables/columns

tamassoltesz:
(#1386)

Calls Storage.verifySchema() once per storage after its first successful initStorage: a mismatch on the base database aborts startup with the missing columns and the SQL to add them; a mismatch on a tenant database is logged at ERROR (SCHEMA_MISMATCH state) and that storage refuses queries until fixed. Deliberately not part of initStorage/pool init, which is re-entered on tenant refresh and lazy pool re-creation.

Also corrects the 12.1.0 changelog: the session_info columns are a manual migration step, not applied automatically.

## Summary of change

## Related issues

- Link to issue1 here
- Link to issue1 here

## Test Plan

## Documentation changes

    - In `build.gradle`

- [ ] Item1
- [ ] Item2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,22 +7,20 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+## [12.2.0]
+
+- **Upgrade note: the core now verifies the database schema at startup and, by default (`schema_check_strict_mode: true`), refuses to start when the base database is missing a manual migration — run the migration SQL from the CHANGELOGs (or set `schema_check_strict_mode: false`) before upgrading**
+- Verifies the database schema at startup: any database (base or tenant) missing a table or column this version needs is reported with the missing columns plus the SQL to add them
+- New config `schema_check_strict_mode` (boolean, default `true`, config.yaml / env only): in strict mode a mismatched tenant database refuses all queries until the migration is applied — re-checked every minute, so it resumes within a minute of the migration, no restart needed; the core still boots and serves all other tenants
+- With `schema_check_strict_mode: false`, mismatches are only logged: everything keeps working and just the queries touching the missing schema fail, with a "Schema mismatch ... check the core error logs" hint instead of a raw SQL error
+- Corrects the 12.1.0 migration note: the `session_info` columns are a manual step, not applied automatically
+
 ## [12.1.2]
 
 - Bulk import no longer borrows from the live connection pool: each worker claims, imports and finalises its chunk on one connection from a dedicated pool sized to `bulk_migration_parallelism`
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,16 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Bulk import keeps claimed `bulk_import_users` rows locked until they are deleted or error-marked; a failed chunk rolls back to a savepoint instead of releasing the claim
 - Bulk import deletes only the rows of the partition it imported and bounds immediate retries after a database rollback
 - `BulkImport.importUser` (the single-user import API) uses the same dedicated-pool mechanism instead of a throwaway pool per call
+- Verifies the database schema at startup: any database (base or tenant) missing a table or column this version
+  needs is reported with the missing columns plus the SQL to add them
+- **New config `schema_check_strict_mode` (boolean, default `true`, config.yaml / env only). In strict mode the
```
porcellus: I think this should be a new minor version instead, and this should be moved to the top (or a one liner version of this)

CHANGELOG.md
```diff
@@ -13,6 +13,16 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
...
+  core REFUSES TO START when the base database schema is out of date (a skipped manual migration), and tenant
+  databases with a mismatched schema are not served until the migration is applied. If your database schema is
```
porcellus: We may want to note that it requires a restart (or fix the need for it)

porcellus (changes requested):

CHANGELOG.md
tamassoltesz: Agreed on both — done. This is now 12.2.0 (`build.gradle` bumped) with its own changelog section, leaving 12.1.2 to just the bulk-import fixes from #1385, and the section now leads with a one-liner upgrade warning in bold, with the details as normal bullets below it. Once #1385 merges I'll re-target this to a 12.2 branch per the release checklist.

CHANGELOG.md
tamassoltesz: Good catch — as written it was misleading: a fenced tenant storage was only re-verified when its storage got reloaded (a tenant config/list change or a restart), because the config-sync cron short-circuits when nothing changed. Went with "fix the need for it": the cron (`SyncCoreConfigWithDb`, every 60s) now re-runs `verifySchema` on loaded storages after its refresh — a verified storage is a cached no-op (no query, just a volatile read), a still-mismatched one costs a single `information_schema` query per tick and stays quiet since the startup ERROR already carries the report. So a strict-mode tenant storage resumes within a minute of the migration SQL being applied, no restart needed; the base database in strict mode still (intentionally) blocks startup. The plugin logs an INFO line on the transition, and the strict-mode tenant test now recovers via the cron instead of a manual `loadStorageLayer()`. Changelog reworded accordingly.

PR #1409, opened 2026-09-01
fix: enforce dashboard session expiry at verification time (GHSA-w9fp-wv6g-pqv8)

tamassoltesz:
## Summary

Fixes the private advisory **GHSA-w9fp-wv6g-pqv8** (CWE-613, insufficient session
expiration). `Dashboard.isValidUserSession` treated "a row exists for this session id" as
"the session is valid" — it checked user suspension and nothing else, never consulting the
`expiry` column that the same code persists. `getEmailFromSessionId` had the same gap. The
only thing ending a dashboard session was the `DeleteExpiredDashboardSessions` cron, whose
first sweep runs 12 hours after process start, so a core that restarts more often than every
12 hours never ran it, and even a long-lived process left a 0–12 h window past the 30-day
lifetime.

Net effect: an expired dashboard session token kept authenticating as the dashboard
administrator — full admin control of the user store (list/search users, read/modify
account info, change email/password, delete users, read/modify tenant core config).

I verified the report against source at `v12.1.1` and this `12.2` base: the two methods and
`getSessionInfoWithSessionId` are byte-for-byte the unfixed version on both.

## Fix

Enforce the lifetime where the session is resolved, so it holds for every storage backend at
once (the check is above the storage query, in storage-agnostic domain code):

- `isValidUserSession` — an expired session is rejected **and revoked** on access, so it stops
  authenticating immediately regardless of whether the cleanup cron has run.
- `getEmailFromSessionId` — returns `null` for an expired session (defence in depth; it is a
  public method and a second read path to the same row).

The cron is unchanged; it remains useful for reclaiming rows in bulk, it is just no longer the
only thing enforcing the lifetime.

## Test

`DashboardTest.testExpiredSessionIsRejectedAtVerification` reproduces the advisory: a session
row written with a past `expiry` (no waiting on the cron) must not authenticate
(`isValidUserSession` → false), must not leak the email (`getEmailFromSessionId` → null), and
is revoked on access; a freshly issued session is the positive control. Fails on the base,
passes with this change.

Ran `io.supertokens.test.dashboard.DashboardTest` against PostgreSQL locally — 7/7 pass.

## Follow-up (not in this PR)

Defence in depth: add `AND expiry > ?` to `getSessionInfoWithSessionId` in each storage
implementation (postgresql / mysql / in-memory) so the invariant holds even if a future
caller forgets the verification-time check. Left out here to keep the security fix to one repo
and reviewable; happy to open the storage-side PRs if you'd like them alongside.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+- Dashboard session tokens are now checked for expiry when verified: `Dashboard.isValidUserSession` and `getEmailFromSessionId` reject a session whose `expiry` has passed (and revoke the expired row) instead of relying solely on the twice-daily cleanup cron, which does not run for the first 12 hours of a process' life.
+
 ## [12.2.0]
 
 - Fixes app and connection URI domain configuration updates being incorrectly rejected as conflicting when affected
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
-- Dashboard session tokens are now checked for expiry when verified: `Dashboard.isValidUserSession` and `getEmailFromSessionId` reject a session whose `expiry` has passed (and revoke the expired row) instead of relying solely on the twice-daily cleanup cron, which does not run for the first 12 hours of a process' life.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+- Dashboard session tokens are now checked for expiry when verified: `Dashboard.isValidUserSession` and `getEmailFromSessionId` reject a session whose `expiry` has passed (and revoke the expired row) instead of relying solely on the twice-daily cleanup cron, which does not run for the first 12 hours of a process' life.
```
tamassoltesz: keep this a one liner

PR #1422, opened 2026-09-10
fix: FeatureFlag resource resilience & non-blocking license sync (PLAN-013)

tamassoltesz:
Base branch for PLAN-013 units 1-3. Cut from v12.3.0-canary; targets the 12.3 release line. Sub-PRs (units 1-3, independent) merge here; this merges to 12.3 and ships in 12.3.0. No migrations, no version bump.

Plan: PLAN-013

https://claude.ai/code/session_01CTU7EfsTCdJKbdbRjigTiN


CHANGELOG.md
```diff
@@ -5,13 +5,19 @@ All notable changes to this project will be documented in this file.
...
 - Active-user activity and its projection now live on the user's own storage (not the app-public tenant), and active-user counts (`GET /users/count/active`, telemetry MAU) sum across every storage backing the app, so a user on a tenant with its own database is counted correctly.
 - In-memory (SQLite) parity for the ledger and rollup storage contract (transactional audit insert, last-active fold/reconcile, windowed event read, and the connection-taking sign-up/tenant-removal variants).
+- `supertokens_saas_load_only_cud` now normalizes the configured value and each connectionUriDomain the same way before comparing, so a CUD that differs only by case/port is no longer silently dropped (which previously wiped its resources and surfaced as a log-free tenant-not-found); dropping a previously-loaded CUD is now logged.
+- The EE license-key sync no longer runs synchronously in the `FeatureFlag` constructor (on the boot thread and under the resource-distributor lock on every tenant/app reload); it now runs on the `EELicenseCheck` cron with a short jittered initial delay after startup, then daily as before. Enabled features are served from the last persisted sync in the meantime.
```
tamassoltesz: these should be also under 12.3.1

tamassoltesz (commented):
Reviewed all three units for correctness, layering and security. Compiled the branch locally and ran the changed test classes; the results turned up one blocking issue.

**Blocking – three EE tests still assume the boot-time sync and fail on this branch** (Postgres, local run; CI skipped `Tests` on all three sub-PRs and is still pending here):

- `ee/.../EETest.testThatInvalidOpaqueKeyOnCoreStartCausesNoFeaturesToBeLoaded` (line ~1101): waits 15s for `LICENSE_KEY_CHECK_NETWORK_CALL` after `STARTED`, but the cron's delay is unpinned (30–60s) → times out; gave up after 3 retries (52s).
- `ee/.../EETest.testThatInvalidstatelessKeyOnCoreStartCausesNoFeaturesToBeLoaded` (line ~1148): asserts `getEnabledFeatures().length == 0` straight after restart, but the DB still holds `[TEST]` until the cron runs → `expected:<1> but was:<0>`.
- `ee/.../api/SetLicenseKeyAPITest.testCallingAPIToSyncLicenseKey` (line 131): `assertFalse(getIsLicenseKeyPresent())` NPEs – `EEFeatureFlag.isLicenseKeyPresent` is `null` until the first sync, and that used to happen in the constructor. Either the test should accept `null`/skip that pre-check, or the field should start as `Boolean.FALSE`; both are fine, just needs a decision.

The first two need the same `CronTaskTest.setIntervalInSeconds(EELicenseCheck.RESOURCE_KEY, 1)` + `startProcess()` + wait-for-event treatment their siblings got in this PR (`startCoreWithInvalid*LicenseKey`). Everything else I ran is green: `MultitenancyHelperFilterTest` 5/5, `FeatureFlagTest` 13/13 (+1 skipped locally because file logging is off in this env), the other 50 EE tests, and `CronjobTest.testThatThereAreTasksOfAllCronTaskClassesAndHaveCorrectIntervals`.

**Security** – nothing negative found:
- The `Main.isTesting` construction hook follows the existing `EE_FOLDER_LOCATION` pattern and is unreachable outside tests.
- Keep-previous can't widen access: a kept `EEFeatureFlag` only reads its own app's `FEATURE_FLAG`/`LICENSE_KEY` rows.
- The delayed validation window at boot is bounded to 60s and serves the last *persisted* sync, never a broader set.
- The request-side `supertokens_saas_load_only_cud` gate (`WebserverAPI:355`) is unchanged and still an exact match on the normalized config value, so the filter normalization can't admit a disallowed CUD on the request path.

**Units** – unit 1 (keep-previous + 5-arg logging + `Logging.error` null fix) is correct and is the part that actually addresses the prod symptom; its regression test exercises both the drop and the null-message path. Two substantive comments inline: unit 3's normalization turns out to be inert for API-created CUDs (only the drop logging carries weight), and unit 2 changes when runtime-created apps first sync. One fragile timing window in `testNetworkCallIsMadeInCoreInit`, plus nits.

For the record, the same message-only `Logging.error(..., e.getMessage(), false)` catch that lost the log here is still used in the sibling `loadForAllTenants` of `AccessTokenSigningKey`, `RefreshTokenKey`, `JWTSigningKey`, `SigningKeys`, `SAMLCertificate` and in `Config.java:150/381`; out of scope for PLAN-013 as re-cut, just noting it.

src/main/java/io/supertokens/multitenancy/MultitenancyHelper.java
```diff
@@ -401,11 +403,92 @@ private TenantConfig[] getFilteredTenantConfigs(TenantConfig[] inputTenantConfig
...
+     * path uses ({@link Utils#normalizeAndValidateConnectionUriDomain}), so comparisons are
+     * normalized-to-normalized. The configured {@code supertokens_saas_load_only_cud} is already
+     * normalized on config load (see {@code CoreConfig}), but a DB-sourced connectionUriDomain read via
+     * {@link TenantIdentifier#getConnectionUriDomain()} is only trimmed and lower-cased, so this closes
+     * the case/port gap between the two operands. (The non-throwing variant only strips case and port;
+     * a scheme or trailing slash hits its internal path check, is caught, and is returned unchanged, but
+     * a stored connectionUriDomain can never carry either because the create path normalizes with the
+     * throwing variant.)
+     */
+    static String normalizeConnectionUriDomainForComparison(String connectionUriDomain) {
```
tamassoltesz: **Unit 3 – the normalization can't change the outcome for any API-created CUD, and a CUD it would newly retain is unreachable anyway.**

Tracing both operands:

- The stored value: `CreateOrUpdateConnectionUriDomainAPI:55` / `V2API:50` run the input through `Utils.normalizeAndValidateConnectionUriDomain(cud)` (throwing variant) before it is persisted. `CoreConfig:1239` runs `supertokens_saas_load_only_cud` through the *same* variant. So for anything created via the API, stored string == config string already.
- `TenantIdentifier.getConnectionUriDomain()` already lower-cases, so the "case" delta this closes never existed. `MultitenancyHelperFilterTest.cudDifferingOnlyByCaseIsRetained` passes on the pre-PR `.equals` code as well, so it isn't a regression test for anything.
- The only remaining delta is a **port** in a stored row, which can only come from an out-of-band DB write. Even then, retaining it doesn't make it usable: `WebserverAPI.getConnectionUriDomain` (line ~352) checks presence against the raw stored set (`dangerous_allCUDsFromDb`) using a port-less `req.getServerName()`, so no request can ever route to `example.com:3567`. Loading it just brings up its storage connections and cron entries for nothing.

So this part of the unit is inert rather than wrong, but the CHANGELOG line ("a CUD that differs only by case/port is no longer silently dropped") and the class javadoc claim a fix for a drop that can't occur through the API. Suggest:

1. Keep the *drop-of-a-previously-loaded-CUD* logging below – that is the genuinely useful part of this unit and does cover the "present, then wiped" symptom.
2. Either drop the normalization (+ the case test) and reword the changelog to just the logging, **or** if we want to be tolerant to out-of-band rows, apply the same normalization in `isConnectionUriDomainPresentInDb` / `dangerous_allCUDsFromDb` too, so the filter and the request gate agree on what "the same CUD" is. Half-normalizing only the filter leaves the system inconsistent.

Nit on the same block: this pulls `io.supertokens.webserver.Utils` + `jakarta.servlet.ServletException` into the multitenancy layer (there's precedent in `CoreConfig`, so not blocking); if the normalizer stays, `io.supertokens.utils.Utils` would be the better home and the `ServletException` catch could go away.

ee/src/main/java/io/supertokens/ee/EEFeatureFlag.java
```diff
@@ -106,18 +106,13 @@ public void constructor(Main main, AppIdentifier appIdentifier) {
...
-            // here.
-        } catch (Throwable ignored) {
-        }
+        // No license sync here on purpose. syncFeatureFlagWithLicenseKey() does synchronous DB stats
+        // queries + an HTTPS call to api.supertokens.com; running it in this constructor meant it ran on
+        // the boot thread (before the webserver was up) and under the global resource-distributor lock on
+        // every tenant/app reload. The EELicenseCheck cron performs the initial sync shortly after boot
+        // (jittered initial delay) and daily thereafter. Enabled features remain available in the meantime
+        // from the DB-persisted value of the last successful sync (getEnabledEEFeaturesFromDbOrCache), so
+        // nothing is lost between boot and the first cron pass.
```
tamassoltesz: **Unit 2 – "nothing is lost between boot and the first cron pass" is true for boot, but apps created/updated *after* boot now wait up to a day for their first sync.**

Before this change the constructor ran for every `new FeatureFlag(main, app)`, i.e. also on `incrementalReloadResources` after a tenant/app create or update, so a new app got an immediate `syncFeatureFlagWithLicenseKey()`. Now `refreshCronjobs()` only adds the app to `EELicenseCheck`'s tenant list, and `Cronjobs.addCronjob` schedules the task exactly once (`tasks.contains`), so the jittered initial delay applies only to the first construction at boot; a runtime-created app's first sync – and the paid-usage stats report that `verifyLicenseKey(rootKey)` sends for it – happens on the next daily tick.

I checked the gating side and it's fine: an absent `FEATURE_FLAG` row reads as `{}` in `getEnabledEEFeaturesFromDbOrCache`, which is exactly what the sync would have written for a key-less app, and `PUT /ee/license` still syncs inline. So this is a stats-latency change, not a feature-gate change. Worth (a) saying so in this comment / the changelog line so nobody reads "nothing is lost" as covering runtime creates, and (b) optionally kicking a one-off `EELicenseCheck` run when `refreshCronjobs()` sees new apps, if same-day stats for new apps matter.

src/test/java/io/supertokens/test/FeatureFlagTest.java
```diff
@@ -901,10 +908,21 @@ public void testNetworkCallIsMadeInCoreInit() throws Exception {
...
+        process = TestingProcessManager.startIsolatedProcess(args, false);
+        // EELicenseCheck.RESOURCE_KEY lives in the ee module, which is not on this (core) test's compile
+        // classpath, so refer to it by its literal resource-key string.
+        CronTaskTest.getInstance(process.getProcess())
+                .setIntervalInSeconds("io.supertokens.ee.cronjobs.EELicenseCheck", 3);
         process.startProcess();
         assertNotNull(process.checkOrWaitForEvent(ProcessState.PROCESS_STATE.STARTED));
+
+        // No synchronous license-server call during boot...
+        assertNull(process.checkOrWaitForEvent(ProcessState.PROCESS_STATE.LICENSE_KEY_CHECK_NETWORK_CALL, 1000));
```
tamassoltesz, line 925: **Fragile timing: the 3s pin and this 1s "no call" window run on different clocks.**

The cron is scheduled inside `loadFeatureFlag` during `Main.init`, so its 3s countdown starts there. Between that and `STARTED` we still run `loadSigningKeys` (RSA keygen on a fresh test DB), ~15 `Cronjobs.addCronjob` registrations and the Tomcat start. Locally that gap is well under a second (test passes here), but the margin is only ~2s and it shrinks on a loaded CI runner or a cold DB; if it is ever exceeded the cron fires inside this window and the `assertNull` fails.

Two related points:

- `CronTaskTest.setIntervalInSeconds` pins the *interval* too (`EELicenseCheck` reads the interval map in both `getIntervalTimeSeconds` and `getInitialWaitTimeSeconds`), so after the first run the cron loops every 3s for the rest of the test. Same for the 1s pins in `EETest`, where each loop is a live license-server call for the opaque keys. `CronTaskTest` already has `setInitialWaitTimeInSeconds`/`getInitialWaitTimeInSeconds` and `BackfillReservationTables`, `ProcessBulkImportUsers` and `RollupUserLastActive` honour it – `EELicenseCheck.getInitialWaitTimeSeconds` should read that one now that the delay is what these tests pin.
- With that seam, pin the delay to something like 10s here (the un-timed `checkOrWaitForEvent` waits 15s, so it still fits) and the boot-vs-window race goes away.

src/main/java/io/supertokens/output/Logging.java
```diff
@@ -205,7 +205,9 @@ public static void error(Main main, TenantIdentifier tenantIdentifier, String er
             // error and continue below.
         }
         try {
-            err = err.trim();
+            // Guard against a null message (e.g. a caller passing e.getMessage() for an exception with
+            // no message). Previously err.trim() threw an NPE that was swallowed below, logging nothing.
+            err = err == null ? "" : err.trim();
```
tamassoltesz: Nit: with a null message this now writes a line whose payload is just the formatted tenant prefix and nothing else. A fixed placeholder (e.g. `"(no message)"`) keeps the null-message fix but makes such lines greppable and makes it obvious at the call site that someone passed `e.getMessage()` of a message-less exception.

ee/src/test/java/io/supertokens/ee/test/CronjobTest.java
```diff
@@ -35,16 +35,25 @@ public void beforeEach() {
...
-                            new KeyValueInfo(EETest.OPAQUE_LICENSE_KEY_WITH_TEST_FEATURE));
+            // Persist an enabled feature set for a valid key so the restarted process below serves it from
+            // the DB (the boot-time constructor no longer syncs). This is skipped on in-memory storage, not
+            // to exclude the test from in-mem runs, but because setLicenseKeyAndSyncFeatures performs a live
+            // license-server sync that would be pointless there: FeatureFlag.getEnabledFeatures() always
+            // returns every feature on in-mem (see FeatureFlag#getEnabledFeatures), so a cron-driven
+            // revocation cannot be observed. The assertions in the next block are therefore already gated to
+            // real SQL storage (the STORAGE_TYPE.SQL && !sqlite guard) exactly as they were before this PR;
+            // on in-mem the block below still starts/stops the process but makes no revocation assertions.
+            if (!StorageLayer.isInMemDb(process.main)) {
```
tamassoltesz: Nit: this identical 8-line comment + guarded `setLicenseKeyAndSyncFeatures` block is pasted four times in this file. A small helper (`persistLicenseKeyIfSqlStorage(process, key)`) with the comment once would keep the four tests readable.

ee/src/test/java/io/supertokens/ee/test/EETest.java
```diff
@@ -183,15 +185,23 @@ public void testNetworkCallMadeOnCoreStartIfLicenseKeyPresent() throws Exception
             TestingProcessManager.TestingProcess process = TestingProcessManager.start(args, false);
             Assert.assertNull(
                     process.checkOrWaitForEvent(ProcessState.PROCESS_STATE.LICENSE_KEY_CHECK_NETWORK_CALL, 1000));
+            // The license sync no longer runs synchronously in EEFeatureFlag's constructor on the boot
+            // thread; the EELicenseCheck cron performs it shortly after startup. Pin its (jittered) initial
+            // delay to 1s so this is deterministic.
+            CronTaskTest.getInstance(process.main).setIntervalInSeconds(EELicenseCheck.RESOURCE_KEY, 1);
```
tamassoltesz, line 191: Nit: `testNetworkCallMadeOnCoreStartIfLicenseKeyPresent` (and `FeatureFlagTest.testNetworkCallIsMadeInCoreInit`) now assert the opposite of their names – the call is explicitly *not* made on core start any more. Worth renaming so a future reader of a failure doesn't get misled.

PR #1426, opened 2026-09-10
feat: isolated admin connector + DB-free /livez liveness (PLAN-014)

tamassoltesz:
Base branch for PLAN-014 Stage 1. Cut from v12.3.0-canary; targets the 12.3 release line. Sub-PRs (connector + route gate, then /livez and the global request-stats endpoint stacked on it) merge here; this merges to 12.3 and ships in 12.3.0. No migrations, no version bump, no CDI bump.

Plan: PLAN-014

https://claude.ai/code/session_01CTU7EfsTCdJKbdbRjigTiN


CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
...
+- Adds optional per-path route-scope overrides for the admin connector: `admin_only_paths` and `admin_preferred_paths` (comma-separated API paths, yaml/env) layer over the compiled-in `RouteScope` defaults, so an operator can promote a route to `ADMIN_ONLY` or widen one to `ADMIN_PREFERRED` without a code change. Both require `admin_port`, reject overlaps, and fail startup on a path that matches no known API.
+- Adds an admin-only `GET /global-request-stats` endpoint returning process-global, in-memory request counts bucketed by response-status class (`2xx`/`4xx`/`5xx`), a `total`, and a `since` timestamp (cumulative since process start, never persisted). Additive: the per-app `GET /requests/stats` is unchanged.
+- Adds a DB-free `/livez` liveness endpoint (`ADMIN_ONLY`, no api-key, no cdi-version) for the container health check: it does no storage access and takes no shared lock, so it stays responsive on the admin connector's isolated pool under data-plane thread saturation and DB-pool exhaustion (unlike `/hello`, which does a storage round-trip and is a readiness check).
```
tamassoltesz, line 14: these should be under 12.3.1

tamassoltesz (commented):
also, missing the build version change from build.gradle

src/test/java/io/supertokens/test/AdminConnectorTest.java
```diff
@@ -0,0 +1,538 @@
...
+
+import static org.junit.Assert.*;
+
+/**
+ * Tests for the opt-in admin connector (a second Tomcat connector on a separate port with its own thread pool)
+ * and the port-scoped route classification enforced by PathRouter. See issue #1427 / PLAN-014.
+ */
+public class AdminConnectorTest {
+
+    private static final int ADMIN_PORT = 3599;
```
tamassoltesz: **Flaky-in-CI risk: fixed admin port under parallel test forks.**

`build.gradle` runs test classes with `maxParallelForks = availableProcessors`. The harness copes with that for the *main* port (isolated processes take `getFreePort()`, and `setValueInConfig` writes a per-worker `config<workerId>.yaml`), but the admin port is hard-coded: `3599` here **and** in `LivezTest`, `3598` in `GlobalRequestStatsTest`. If two forks pick up `AdminConnectorTest` and `LivezTest` at the same time, the second Tomcat can't bind `3599`, `start()` throws `QuitProgramException`, and every test in that class fails with "another instance is already running on the same port". It won't reproduce locally when the classes run one at a time (they pass for me), which is exactly the kind of failure that shows up only on a busy CI runner.

Suggest allocating the admin port per test with `TestingProcessManager.getFreePort()` (and building `ADMIN` from it) instead of the constants, in all three classes.

src/main/java/io/supertokens/webserver/api/core/GlobalRequestStatsAPI.java
```diff
@@ -0,0 +1,70 @@
...
+import io.supertokens.webserver.WebserverAPI;
+import jakarta.servlet.ServletException;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
+
+import java.io.IOException;
+
+/**
+ * Admin-only endpoint exposing the process-global, per-status-class request counts held by
+ * {@link GlobalRequestStats}. Served only on the admin connector ({@link RouteScope#ADMIN_ONLY}); returns 404
```
tamassoltesz: Doc accuracy: "returns 404 on the main port" is only true when `admin_port` is set. With the default single-connector topology (which `LivezTest.testLivezReturns200OnMainPortWhenAdminPortUnset` deliberately pins) the gate is inert and this endpoint is served on the data-plane port, guarded only by the api-key check (i.e. open on a core with no `api_keys`). That is consistent with the design ("exactly as before" when disabled), but the class doc and the `RouteScope` javadoc should say it explicitly so nobody reads `ADMIN_ONLY` as a hard guarantee. Same applies to the `LivezAPI` header comment.

config.yaml
```diff
@@ -107,6 +107,29 @@ core_config_version: 0
 # max_server_pool_size:
 
 
+# (OPTIONAL | Default: null) integer value. The port for a second webserver connector that serves liveness and
+# control-plane routes from a thread pool separate from the data-plane pool. When set to null, the admin connector
+# is disabled and all routes are served on the main port exactly as before.
+# admin_port:
```
tamassoltesz, line 117: Docs/security: the admin port carries `/ee/license` (PUT, sets the license key) and `/global-request-stats`, and it applies the same `api_keys` / IP rules as the main port – there is no separate credential for it. Worth one sentence here that the admin port is meant for the orchestrator/control plane and should not be exposed beyond the network boundary the health check and control plane live in (and that `host` applies to both connectors). Operators reading "liveness" tend to assume the port is inert.

src/main/java/io/supertokens/config/CoreConfig.java
```diff
@@ -779,6 +814,60 @@ public int getMaxThreadPoolSize() {
...
+                paths.add(p);
+            }
+        }
+        return paths;
+    }
+
+    // Normalize an API path for route-scope override matching: trim, lowercase, ensure a single leading slash, and
+    // drop a trailing slash. PathRouter derives the override key from a matched API's getPath() through this same
+    // method, so the two representations stay in lockstep.
+    public static String normalizeApiPath(String path) {
```
tamassoltesz, line 857: Nit (SRP): API-path normalization for route matching is a webserver concern; having `CoreConfig` own it (and `PathRouter` reach back into the config class for it) inverts the dependency direction. `CoreConfig` only needs it to detect the admin_only/admin_preferred overlap, and that check could just as well live next to `validateRouteScopeOverrides()` in `PathRouter`, which already has the registered API list. Not blocking.

src/main/java/io/supertokens/webserver/PathRouter.java
```diff
@@ -17,23 +17,104 @@
...
 
 public class PathRouter extends WebserverAPI {
     private static final long serialVersionUID = -3199188474453935983L;
 
     List<WebserverAPI> apis;
 
+    // Snapshot of the admin-connector routing state (enabled flag + admin port). The base config is immutable for
+    // the lifetime of the process — the admin port cannot change without a restart, and a restart builds a fresh
+    // Webserver (hence a fresh PathRouter). So we read it from the ResourceDistributor exactly once and reuse it on
+    // every request instead of taking the read lock on the config lookup on the hot path. Computed lazily on first
```
tamassoltesz: Nit: the comment's justification is off – `ResourceDistributor.getResource` is a `ConcurrentHashMap` lookup with no read lock (only `withResourceDistributorLock` is `synchronized`, and that's the reload path). Caching the gate is still fine (saves the `Config.getInstance` indirection and the override-map rebuild per request); just drop the lock rationale so the next reader doesn't go looking for one.

src/main/java/io/supertokens/webserver/WebserverAPI.java
```diff
@@ -598,6 +622,10 @@ protected void service(HttpServletRequest req, HttpServletResponse resp) throws
                 false);
 
         if (tenantIdentifier != null) {
+            // Process-global, per-status-class counts (across all apps). Independent of the per-app RequestStats
+            // below; keyed at the base tenant so it never throws for an unknown app. Requests that fail tenant
+            // resolution (tenantIdentifier == null) are intentionally not counted here.
+            GlobalRequestStats.getInstance(main).incrementForStatus(resp.getStatus());
```
tamassoltesz, line 632: Small counting gap worth a line in the `GlobalRequestStats` javadoc: besides the documented `tenantIdentifier == null` case, the `PathRouter` port-gate 404s are sent before this hook and are not counted either. Harmless for the intended use, but if someone compares `total` against the ALB request count for the admin port it'll be off by the misrouted probes.

tamassoltesz (commented):
Reviewed all three units (connector + route gate, `/livez`, global request stats) for correctness, layering and security. Built the branch locally and ran the new test classes plus the config-annotation tests: `AdminConnectorTest` 16/16, `LivezTest` 6/6, `GlobalRequestStatsTest` 4/4, `ConfigTest.testAllConfigFieldsAreAnnotated` green (Postgres). Note CI `Tests` was skipped on all three sub-PRs (#1433-#1435) and is still pending here, so this local run is the first time these have executed outside the authoring agent.

**Correctness** – the core design holds up:
- `req.getLocalPort()` is the right discriminator (socket-level, can't be spoofed via `Host`), and the gate sits after `getAPIThatMatchesPath` so `/appid-x/<tenant>/livez` resolves correctly on either port.
- I traced `/livez` through `WebserverAPI.service`: tenant resolution, `checkIPAccess`, config lookup and the stats hook are all in-memory (`ConcurrentHashMap` in `ResourceDistributor`, the volatile CUD set in `MultitenancyHelper`), so the "no storage, no shared lock" claim is accurate; the only I/O left is the two `Logging.info` lines. `testLivezIsDbFreeWhenStorageIsDown` backs it up.
- `GlobalRequestStats` lives on `BASE_TENANT`; nothing in the reload paths wipes base-tenant resources by anything other than key (`clearAllResourcesWithResourceKey`), so the counters survive tenant refreshes. The hook runs after the catch block, so exception-driven 5xx are counted.
- The boxed-`Integer` env-var fix is real and `admin_port`/`bulk_migration_sleep_between_rounds_in_batch_ms` are the only two boxed fields, so it's complete.
- `admin_port` validation (range, `!= port`, pool size) and the override validation (require `admin_port`, no overlap, must match a registered API, fail at startup) are all covered by tests.

**Security** – nothing negative. `/livez` is credential-free but returns nothing beyond 200 (same posture as `/hello`). `/global-request-stats` and the `/ee/*` routes keep the api-key requirement on both ports. The one thing I'd make explicit in the docs is that the admin port is *not* a separate trust boundary (same `api_keys`, same IP rules, same `host`) – see the inline comment on `config.yaml`. HTTPS is not configured on either connector today, so there's no asymmetry introduced.

**Blocking-ish** – one inline comment on the tests: the three new test classes hard-code admin ports (3598/3599) while `build.gradle` runs classes in parallel forks; that will bite in CI even though it passes when run one class at a time. The rest are doc-accuracy and nits.

Your earlier comment on the changelog/version bump has been addressed in 6a65ae83; not repeating it.

